PR #27, opened 2021-05-05
T941 refactor file cell

Ambrdctr:


flowbe (changes requested):
Highlight of grid cell with thumbnail not good

flowbe (approved):

PR #63, opened 2021-05-19
LoadingIndicator added to fileDetail

Ambrdctr:
Signed-off-by: Ambroise Decouttere <ambroise.decouttere@infomaniak.com>


PhilippeWeidmann (changes requested):
Loading indicator only when loading for more than 0.5 sec (as done in the filelist)
Side note:
`getShareListFor` is called twice:
- Once in viewDidLoad
- Once in viewDidAppear

We should prevent this on initial loading

PR #76, opened 2021-05-31
Improved picture view

PhilippeWeidmann:


kDrive/UI/Controller/Menu/PhotoListViewController.swift
```diff
@@ -231,6 +330,6 @@ extension PhotoListViewController: UICollectionViewDelegateFlowLayout {
         }
         let width = collectionView.frame.width - collectionViewLayout.minimumInteritemSpacing * CGFloat(numberOfColumns - 1)
         let cellWidth = width / CGFloat(numberOfColumns)
-        return CGSize(width: cellWidth, height: cellWidth)
+        return CGSize(width: Int(cellWidth), height: Int(cellWidth))
```
flowbe, line 333: Maybe `floor` would be easier to understand

flowbe (changes requested):
Some UI bugs with the headers (left margins, transitions, etc)

flowbe (approved):

PR #179, opened 2021-07-06
Bulk actions and select all (MQ Service)

PhilippeWeidmann:
What's left:
Translations:
- [x] KDriveStrings.Localizable.buttonDeselectAll
- [x] Progress notification delete -> start, progress, end 
- [x] Progress notification move -> start, progress, end 

Testing:
- [x] What happens if we select unselectable items no rights, other), should we care ? (
- [x] What happens if I switch account regarding the IPS token
- [x] What happens if I lose network access ?


flowbe:
We should also handle cancellation

Lnamw (approved):

PR #444, opened 2021-12-02
Folder color

Lnamw:


flowbe (changes requested):
- [x] Increment Realm schema version number
- [x] Fix floating panel size after rotation
- [x] Clean enum names in information floating panels
- [ ] Add unit test for `DriveApiFetcher.updateFolderColor`

flowbe (approved):

PR #783, opened 2022-04-27
Cancel task before creating background task

PhilippeWeidmann:
When rescheduling an upload due to the background task expiration, the uploaded file was renamed (file -> file (1)) because upload task wasn't canceled.


adrien-coye:
Discussed with @PhilippeWeidmann, this MR is deprecated by the work I'm doing with APIV2 / upload in chunks.

PR #953, opened 2022-12-13
Fix categories

### LINKED ISSUE #930: Bug in categories displaying when coming back from the creating category screen
When selecting several categories, then trying to create a new one, only the first previously selected category stays displayed as added (even though the other are added on the file)
Furthemore, the newly created category is not added to the file

valentinperignon:
Fixes #930

### DIFF
--- a/kDrive/UI/Controller/Files/Categories/EditCategoryViewController.swift
+++ b/kDrive/UI/Controller/Files/Categories/EditCategoryViewController.swift
@@ -26,6 +26,7 @@ class EditCategoryViewController: UITableViewController {
     var category: kDriveCore.Category?
     /// The file to add the category to after creating it.
     var fileToAdd: File?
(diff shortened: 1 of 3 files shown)


kDrive/UI/Controller/Files/Categories/EditCategoryViewController.swift
```diff
@@ -26,6 +26,7 @@ class EditCategoryViewController: UITableViewController {
     var category: kDriveCore.Category?
     /// The file to add the category to after creating it.
     var fileToAdd: File?
```
PhilippeWeidmann, line 28: Remove me

valentinperignon: You've been removed

Ambrdctr (approved):

PR #959, opened 2023-01-03
Update "empty folder" strategy

valentinperignon:


adrien-coye (approved):
@valentinperignon explained to me what this MR is fixing.
I ran the code on the test iPhone 8. It works as expected.
I also looked at the code, it looks coherent with what we try to fix.

PhilippeWeidmann (approved):

PR #960, opened 2023-01-04
Fix JPEG data

valentinperignon:
Keep creation and modification date

### DIFF
--- a/kDriveCore/Utils/PHAsset+Exension.swift
+++ b/kDriveCore/Utils/PHAsset+Exension.swift
@@ -79,13 +79,15 @@ extension PHAsset {
 
         let targetURL = FileImportHelper.instance.generateImportURL(for: resourceUTI)
         do {
-            if shouldTransformIntoJPEG {
-                if try await writeJpegData(to: targetURL, resource: resource, options: requestResourceOption) {
-                    return targetURL
-                }
+            guard shouldTransformIntoJPEG else {
+                try await PHAssetResourceManager.default().writeData(for: resource, toFile: targetURL, options: requestResourceOption)
+                return targetURL
+            }
+
+            guard try await writeJpegData(to: targetURL, resource: resource, options: requestResourceOption) else {
                 return nil
             }
-            try await PHAssetResourceManager.default().writeData(for: resource, toFile: targetURL, options: requestResourceOption)
+
             return targetURL
         } catch {
             let breadcrumb = Breadcrumb(level: .error, category: "PHAsset request data and write")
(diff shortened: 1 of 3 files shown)


kDriveCore/Utils/PHAsset+Exension.swift
```diff
@@ -96,14 +95,27 @@ extension PHAsset {
...
+    private func getJpegData(for resource: PHAssetResource, options: PHAssetResourceRequestOptions) async throws -> Data? {
         return try await withCheckedThrowingContinuation { continuation in
             var imageData = Data()
-            PHAssetResourceManager.default().requestData(for: resource, options: requestResourceOption) { data in
+            PHAssetResourceManager.default().requestData(for: resource, options: options) { data in
                 // Get all pieces of data
                 imageData.append(data)
             } completionHandler: { error in
-                if let error = error {
+                if let error {
```
adrien-coye, line 118: That's cool we no longer need the `error = error`

kDriveCore/Utils/PHAsset+Exension.swift
```diff
@@ -80,8 +80,7 @@ extension PHAsset {
         let targetURL = FileImportHelper.instance.generateImportURL(for: resourceUTI)
         do {
             if shouldTransformIntoJPEG {
```
adrien-coye, line 82: I think we can use guards to achieve the same with less nested ifs

```
do {
            guard shouldTransformIntoJPEG == true else {
                try await PHAssetResourceManager.default().writeData(for: resource, toFile: targetURL, options: requestResourceOption)
                return targetURL
            }
            
            guard let jpegData = try await getJpegData(for: resource, requestResourceOption: requestResourceOption) else {
                return nil
            }
            
            try jpegData.write(to: targetURL)
            return targetURL
        }
```

kDriveCore/Utils/PHAsset+Exension.swift
```diff
@@ -96,14 +95,27 @@ extension PHAsset {
         return nil
     }
 
-    private func getJpegData(for resource: PHAssetResource, requestResourceOption: PHAssetResourceRequestOptions) async throws -> Data? {
+    private func writeJpegData(to url: URL, resource: PHAssetResource, options: PHAssetResourceRequestOptions) async throws -> Bool {
+        if let jpegData = try await getJpegData(for: resource, options: options) {
```
adrien-coye, line 99: We can use a guard here also I think

adrien-coye (approved):
All good 👍

PhilippeWeidmann (approved):

PR #971, opened 2023-03-08
feat: Make `with=` type safe

PhilippeWeidmann:


adrien-coye (approved):
Okay, I got the way to encapsulate the query params.

PR #976, opened 2023-03-13
Fix Files.app extension

adrien-coye:
- Fix Realm using more than 20MiB in Files.app.
- Tune upload mechanism to use ~4MiB max during uploads in Files.app.
- Fix async issues resulting in inconsistencies in Files.app.
- Disabled some notifications when uploading from Files.app.

### DIFF
--- a/kDriveCore/Data/Cache/DriveFileManager.swift
+++ b/kDriveCore/Data/Cache/DriveFileManager.swift
@@ -79,12 +79,7 @@ public class DriveFileManager {
                     }
                 }
 
-                // Migration for Upload APIV2
-                if oldSchemaVersion < 13 {
-                    migration.deleteData(forType: UploadFile.className())
-                }
-
-                // Migration for Upload APIV2 build3
+                // Migration for Upload With Chunks
                 if oldSchemaVersion < 14 {
                     migration.deleteData(forType: UploadFile.className())
                 }


kDriveFileProvider/FileProviderExtension+Actions.swift
```diff
@@ -20,43 +20,60 @@ import FileProvider
...
 
-        Task {
             do {
-                let response = try await driveFileManager.apiFetcher.deleteDefinitely(file: ProxyFile(driveId: driveFileManager.drive.id, id: fileId))
+                let response = try await self.driveFileManager.apiFetcher
+                    .deleteDefinitely(file: ProxyFile(driveId: self.driveFileManager.drive.id, id: fileId))
                 if response {
-                    FileProviderExtensionState.shared.workingSet.removeValue(forKey: itemIdentifier)
+                    self.fileProviderState.removeWorkingDocument(forKey: itemIdentifier)
                     self.manager.signalEnumerator(for: .workingSet) { _ in }
```
adrien-coye, line 77: switch to async/await

kDriveCore/Data/Cache/DriveFileManager.swift
```diff
@@ -69,6 +83,11 @@ public class DriveFileManager {
                 if oldSchemaVersion < 13 {
                     migration.deleteData(forType: UploadFile.className())
                 }
+
+                // Migration for Upload APIV2 build3
+                if oldSchemaVersion < 14 {
```
PhilippeWeidmann, line 88: Since we are doing the same for 13 and 14 we can simply remove the block for 13

adrien-coye: nice catch

kDriveCore/FileProvider/FileProviderItem.swift
```diff
@@ -19,9 +19,8 @@
 import CoreServices
 import FileProvider
 import InfomaniakDI
-import kDriveCore
 
-extension NSFileProviderItemIdentifier {
+public extension NSFileProviderItemIdentifier {
```
PhilippeWeidmann, line 23: Why do we need it to be public ?

PR #977, opened 2023-03-17
Direct Upload for Empty files

adrien-coye:


PhilippeWeidmann:
We need to use API V2 also for one shot. For now test with api V1

PhilippeWeidmann (approved):

PR #1001, opened 2023-06-26
Prevent loop in cancel session

adrien-coye:
- A loop may occur when a session is not valid, we try to delete it from the remote, which fails, so we try to remove it …
- Also refactored a bit to differenciate more in the error handling.


kDriveCore/Data/UploadQueue/Operation/UploadOperation.swift
```diff
@@ -436,16 +436,24 @@ public final class UploadOperation: AsynchronousOperation, UploadOperationable,
                 return
             }
 
+            guard remotely == true else {
```
PhilippeWeidmann, line 439:
```suggestion
            guard remotely else {
```

PhilippeWeidmann (approved):

PR #1006, opened 2023-07-19
Bringing back abstracted file sharing into kDrive

adrien-coye:
## Abstract

Bring back the `Core` work related to file sharing in this app.

Also abstracted the on the fly Heic -> Jpg ~[here](https://github.com/Infomaniak/ios-core/pull/53)~

~[use date as default name Core](https://github.com/Infomaniak/ios-core/pull/58)~
~[use date as default name CoreUI](https://github.com/Infomaniak/ios-core-ui/pull/11)~

Shared file name generation between manual import and PhotoLibraryUploader. 

Edited pictures in Photo.app have now a non coliding name.

## Details
- Speedup improvement, scale import workers dynamicaly given a device.
- chore: Bump Core / Core UI

PR #1015, opened 2023-08-04
refactor(PhotoLibraryUploader): split class for readability

adrien-coye:


adrien-coye:
Done here instead https://github.com/Infomaniak/ios-kDrive/pull/1006

PR #1024, opened 2023-09-25
Revert file naming convention

adrien-coye:
- Added more logging for errors presented to the user
- Rollback to the original upload naming scheme
- Unit test upload naming scheme
- Use TolerantDispatchGroup to harden ExpiringActivity

### DIFF
--- a/kDriveCore/Data/Models/DriveError.swift
+++ b/kDriveCore/Data/Models/DriveError.swift
@@ -235,7 +235,7 @@ public struct DriveError: Error, Equatable {
     private static let decoder = JSONDecoder()
 
     /// A specific, not user facing, not localized error
-    internal var underlyingError: Error?
+    var underlyingError: Error?
 
     public init(apiErrorCode: String, httpStatus: Int = 400) {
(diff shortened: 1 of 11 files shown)


kDriveCore/Data/Models/DriveError.swift
```diff
@@ -235,7 +235,7 @@ public struct DriveError: Error, Equatable {
     private static let decoder = JSONDecoder()
 
     /// A specific, not user facing, not localized error
-    private var underlyingError: Error?
+    internal var underlyingError: Error?
```
PhilippeWeidmann, line 238: We don't need `internal` as this is the default access level

adrien-coye: I'm aware, I wanted to be explicit about it as we kind of order parameters given public / private access level. 

I'll remove it asap.

adrien-coye:
For the record: Sonar Cloud is unhappy about code duplication in my unit tests, so not an issue.

PhilippeWeidmann (approved):

PR #1040, opened 2023-10-18
fix: Stopgap to reduce NSInternalInconsistencyException

adrien-coye:
### Abstract

A proposal to try to reduce internal consistency issues.


PhilippeWeidmann:
Will cause flashing on viewDidAppear

adrien-coye:
As discussed, I'll look into [DifferenceKit]](https://github.com/ra1028/DifferenceKit) to fix crashes when one of us has time.

Issue #984, opened 2023-04-12
Investigate upload restarting from beginning when switching network


PhilippeWeidmann:
When user switches from Wifi to 4G the upload starts from scratch instead of keeping uploaded chunks.

adrien-coye:
4.3.4-13 has state negotiation before restarting an upload.

PR #1043, opened 2023-10-24
fix(UploadOperation): Seamless switch between WIFI and Cellular data.

adrien-coye:
Fix https://github.com/Infomaniak/ios-kDrive/issues/984


valentinperignon (approved):
LGTM

PR #1045, opened 2023-10-26
feat: Local albums sync

adrien-coye:


adrien-coye:
Non trivial to implement, requires diffing API, closing for now.

PR #1048, opened 2023-11-02
fix: Update drive register route

PhilippeWeidmann:
Redirect route was changed from drive.infomaniak.com to kdrive.infomaniak.com

### DIFF
--- a/kDrive/UI/Controller/RegisterViewController.swift
+++ b/kDrive/UI/Controller/RegisterViewController.swift
@@ -121,7 +121,7 @@ extension RegisterViewController: WKNavigationDelegate {
         decisionHandler: @escaping (WKNavigationActionPolicy) -> Void
     ) {
         if let host = navigationAction.request.url?.host {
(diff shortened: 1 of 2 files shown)


kDrive/UI/Controller/RegisterViewController.swift
```diff
@@ -121,7 +121,7 @@ extension RegisterViewController: WKNavigationDelegate {
         decisionHandler: @escaping (WKNavigationActionPolicy) -> Void
     ) {
         if let host = navigationAction.request.url?.host {
-            if host == "drive.infomaniak.com" {
+            if host == "kdrive.infomaniak.com" {
```
adrien-coye, line 124: Can we put this in a const file ? (like we do in mail ?)

PhilippeWeidmann: Done

adrien-coye (approved):
all good 👍

PR #1049, opened 2023-11-02
feat: PHAsset revision support

adrien-coye:
### Abstract

An edited picture in the photo roll should be uploaded if sync is enabled.

For technical reasons it will be enabled on __iOS15 and up__ only

### Tech

After discussion, `modifedDate` cannot be trusted, therefore one relyable way to do this is to hash the source image and rely on hashes comparison.

### Dependency

New type in Core to handle hashing a stream of data.
https://github.com/Infomaniak/ios-core/pull/76


kDriveCore/Utils/PHAsset/PHAssetIdentifier.swift
```diff
@@ -0,0 +1,128 @@
...
+        // TODO: Check iCloud behaviour
+        let options = PHAssetResourceRequestOptions()
+        options.isNetworkAccessAllowed = true
+        options.progressHandler = { progress in
+            print("hashing resource \(progress * 100)% …")
+        }
+
+        PHAssetResourceManager.default().requestData(for: bestResource,
+                                                     options: options) { data in
+            // TODO: Hash scanning data, cannot access all at once
```
adrien-coye, line 115: TODO: scan SHA256.

kDriveCore/Utils/PHAsset/PHAsset+Exension.swift
```diff
@@ -23,29 +23,39 @@ import Photos
...
+        return identifier.bestResourceSHA256
     }
 
     // MARK: - Filename
 
     /// Get a filename that can be used by kDrive, taking into consideration the edits that may exists on a PHAsset.
     func getFilename(fileExtension: String,
-                     creationDate: Date? = nil,
-                     modificationDate: Date? = nil,
-                     burstCount: Int? = nil,
```
adrien-coye, line 44: I removed the default parameters, there was a bug with a value not set somewhere it was possible to set the value.

kDriveCore/Utils/PHAsset/PHAssetIdentifier.swift
adrien-coye: Done here https://github.com/Infomaniak/ios-core/pull/76

kDriveCore/Data/UploadQueue/Servicies/PhotoLibraryUploader+Scan.swift
```diff
@@ -206,25 +192,71 @@ public extension PhotoLibraryUploader {
...
+
+        /// Nothing found
+        else {
             return false
         }
+    }
 
-        return true
+    /// Get the current diff algorithm version
+    private func currentDiffAlgorithmVersion() -> Int {
```
PhilippeWeidmann, line 255: This can be a var

PhilippeWeidmann (approved):

PR #1063, opened 2023-11-16
feat: Improved media save error handling

PhilippeWeidmann:


kDriveCore/Utils/SentryDebug.swift
```diff
@@ -240,4 +240,12 @@ public enum SentryDebug {
     public static func captureNoWindow() {
         SentrySDK.capture(message: "Trying to call show with no window")
     }
+
+    // MARK: - Save Media Error
+
+    public static func saveMediaError(_ error: Error) {
```
adrien-coye, line 246: Not to do in this PR, but all code in `SentryDebug` will become async, so it will make sure to not introcude hangs.

adrien-coye (approved):
All good 👍

PR #1066, opened 2023-11-24
chore: Bump i18n

adrien-coye:


adrien-coye:
I'll try from scratch with the dedicated script

PR #1070, opened 2023-11-27
feat: Application shortcuts

Lnamw:


adrien-coye (commented):
Code looks good, I think it would be better with a bit more splitting.

kDrive/AppDelegate.swift
```diff
@@ -231,6 +243,56 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, AccountManagerDeleg
...
+            default:
+                break
+            }
+
+            // reset the shortcut item
+            shortcutItemToProcess = nil
+        }
+    }
+
+    private func openScan(_ mainTabViewController: MainTabViewController, _ driveFileManager: DriveFileManager) {
```
adrien-coye, line 284: I think we can remove the `openScan` and `openPhoto` functions.
And put that code directly line 263 in the `switch shortcutItem.type {`

For now It should be OK, but if you find the function `applicationDidBecomeActive` too long. I guess we can add another file named `AppDelegate+Shortcuts.swift` to keep all the code related to shortcut handling, and keep the `openScan` and `openPhoto` functions there.

What do you think ?

PR #1071, opened 2023-11-29
fix: Review realm objects init and enforce non nullability of primary keys

adrien-coye:
### Tasks

- Enforce non nullability of Primary keys
- Added missing calls to `super.init()` to respect convention.


kDriveCore/Data/Models/Drive/Drive.swift
```diff
@@ -60,13 +60,13 @@ public enum MaintenanceReason: String, PersistableEnum, Codable {
...
-public class DrivePreferences: EmbeddedObject, Codable {
+public final class DrivePreferences: EmbeddedObject, Codable {
     @Persisted public var color = "#0098FF"
     @Persisted public var hide = false
 }
 
-public class Drive: Object, Codable {
-    @Persisted(primaryKey: true) public var objectId = ""
+public final class Drive: Object, Codable {
+    @Persisted(primaryKey: true) public var objectId: String = UUID().uuidString
```
PhilippeWeidmann, line 69: This is not correct, objectId != UUID().uuidString

adrien-coye: Is it less correct than the previous default that was empty string ? `""`

adrien-coye: I'll change this to only set a UUID in the empty init method.

PhilippeWeidmann: I wanted to ensure that we don't have places in the code where we check for .isEmpty on the objectId but this is not the case. So everything seems good 👍

kDriveCore/Data/Models/DriveUser.swift
```diff
@@ -83,8 +83,14 @@ public class DriveUser: Object, Codable, InfomaniakUser {
...
 
+    override public init() {
+        // Required by Realm
+        super.init()
+        // Primary key set as default value
+    }
+    
     public convenience init(user: InfomaniakCore.UserProfile) {
-        self.init()
+        super.init()
```
PhilippeWeidmann, line 93: Doesn't compile

kDriveCore/Data/Models/Drive/Drive.swift
```diff
@@ -60,13 +60,13 @@ public enum MaintenanceReason: String, PersistableEnum, Codable {
...
-public class DrivePreferences: EmbeddedObject, Codable {
+public final class DrivePreferences: EmbeddedObject, Codable {
     @Persisted public var color = "#0098FF"
     @Persisted public var hide = false
 }
 
-public class Drive: Object, Codable {
-    @Persisted(primaryKey: true) public var objectId = ""
+public final class Drive: Object, Codable {
+    @Persisted(primaryKey: true) public var objectId: String
```
PhilippeWeidmann, line 69: Realm doesn't seem happy if the default value isn't provided directly here

adrien-coye: Discussed and modified.

kDriveCore/Data/Cache/PhotoSyncSettings.swift
```diff
@@ -78,11 +79,9 @@ public class PhotoSyncSettings: Object {
...
 
     override public init() {
-        // META: keep SonarCloud happy
-    }
-
-    override public class func primaryKey() -> String? {
-        return "userId"
+        // Required by Realm
+        super.init()
+        self.userId = UUID().uuidString.hashValue
```
adrien-coye, line 84: This can break given the way the property wrapper works.

adrien-coye: Fixed

PR #1075, opened 2023-11-30
fix: Correctly shutdown MQTT client on deinit

PhilippeWeidmann:
MQTT throws a fatal error in case we deinit without shutting down


adrien-coye (approved):
LGTM 👍

PR #1079, opened 2023-12-06
fix: Correctly disconnect user

PhilippeWeidmann:


kDrive/AppDelegate.swift
```diff
@@ -82,6 +83,8 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, AccountManagerDeleg
         let currentState = RootViewControllerState.getCurrentState()
         prepareRootViewController(currentState: currentState)
 
+        (accountManager as? AccountManager)?.delegate = self
```
PhilippeWeidmann, line 86: If I don't cast, i get a crash in the DI. Do you know why @adrien-coye ?

adrien-coye: You cannot declare weak properties in a pure swift protocol (even if inherits from `AnyObject`)
You probably want to stick to a `setDelegate(_ )` method for this case.

You can declare a property weak in a protocol only if @Objc which we probably do not want.

kDrive/AppDelegate.swift
```diff
@@ -285,8 +291,11 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, AccountManagerDeleg
     func refreshCacheData(preload: Bool, isSwitching: Bool) {
         Log.appDelegate("refreshCacheData preload:\(preload) isSwitching:\(preload)")
 
-        @InjectService var accountManager: AccountManageable
-        let currentAccount = accountManager.currentAccount!
+        guard let currentAccount = accountManager.currentAccount else {
+            Log.appDelegate("No account to refresh")
```
adrien-coye, line 295:
```suggestion
            Log.appDelegate("No account to refresh", level: .error)
```

kDrive/AppDelegate.swift
```diff
@@ -458,7 +465,9 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, AccountManagerDeleg
     // MARK: - Account manager delegate
 
     func currentAccountNeedsAuthentication() {
-        setRootViewController(SwitchUserViewController.instantiateInNavigationController())
+        DispatchQueue.main.async { [weak self] in
```
adrien-coye, line 468: Can we use `Task { @MainActor _ `instead ?

adrien-coye (approved):
Code as is looks good. Two remarks if you agree with me, you can make two small changes.

adrien-coye (approved):

PR #1084, opened 2023-12-14
feat: Dedicated test target for online API tests

adrien-coye:
- Can be performed with `tuist test kDriveAPITests` as they are not part of the app.
- To be done later: Update CI to perform them in a dedicated "merge requirement".


.tuist-version
```diff
@@ -1 +1 @@
-3.16.0
\ No newline at end of file
+3.35.5
```
PhilippeWeidmann, line 1: You need to do `tuist bundle` after upgrading version

PhilippeWeidmann (approved):

PR #1087, opened 2023-12-19
feat: Receive file from kmail app

Ambrdctr:


kDrive/AppDelegate.swift
```diff
@@ -221,6 +221,19 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, AccountManagerDeleg
                 StorePresenter.showStore(from: viewController, driveFileManager: driveFileManager)
             }
             return true
+        } else if components.host == "file",
```
adrien-coye, line 224: Nothing really, but I would prefer to use a switch on `components.host`. It's a subjective thing. As you prefer.

kDrive/AppDelegate.swift
```diff
@@ -221,6 +221,19 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, AccountManagerDeleg
                 StorePresenter.showStore(from: viewController, driveFileManager: driveFileManager)
```
adrien-coye, line 221: All the code related to the parsing of `deeplinks` (URL components code here) should be split in a dedicated type that lives in drive-core. That way the app delegate stays clean.

Then we should think about abstracting it and sharing with drive, so we have the benefits of this for all our apps.

adrien-coye (commented):
Okay, just one main thing regarding keeping code outside of the app delegate.

kDrive/AppDelegate.swift
adrien-coye: discussed, not helpful here.

kDrive/Utils/FileActionsHelper.swift
```diff
@@ -469,7 +469,8 @@ public final class FileActionsHelper {
             floatingPanelViewController?.rightButton.isEnabled = driveFileManager.drive.accountAdmin
             floatingPanelViewController?.actionHandler = { _ in
                 driveFloatingPanelController.dismiss(animated: true) {
-                    StorePresenter.showStore(from: viewController, driveFileManager: driveFileManager)
+                    @LazyInjectService var navigationManager: NavigationManageable
```
adrien-coye, line 472: It is a tiny thing really, but by convention we should only use `LazyInjectService` when used as a property.

adrien-coye:
I did a review, it looks good to me. Philippe will give the final go.

kDriveCore/Utils/DeeplinkParser.swift
```diff
@@ -0,0 +1,68 @@
...
+import InfomaniakDI
+import SwiftUI
+
+public struct DeeplinkParser {
+    @LazyInjectService var accountManager: AccountManageable
+    @LazyInjectService var navigationManager: NavigationManageable
+
+    private let window: UIWindow?
+
+    public init(window: UIWindow?) {
```
adrien-coye, line 28: Having a `DeeplinkParser` that requires a reference to UIKit elements seems strange.
Since we do not have an abstracted navigation, I guess we cannot change this without refactoring the navigationManager.

Project.swift
```diff
@@ -160,7 +160,7 @@ let project = Project(name: "kDrive",
                                  bundleId: "com.infomaniak.drive.FileProvider",
                                  deploymentTarget: Constants.deploymentTarget,
                                  infoPlist: .file(path: "kDriveFileProvider/Info.plist"),
-                                 sources: "kDriveFileProvider/**",
+                                 sources: ["kDriveFileProvider/**", "kDrive/Utils/AppFactoryService.swift", "kDrive/Utils/NavigationManager.swift"],
```
PhilippeWeidmann, line 163: This looks weird, why is this needed ?

PhilippeWeidmann (dismissed):

kDriveCore/Utils/DeeplinkParser.swift
adrien-coye: I think we can solve this with a new protocol that can provide the `rootViewController` instead of passing the window as an argument. That way the code would be more abstract, what to you think @Ambrdctr  ?

adrien-coye: Or we should move the view related code to the `navigationManager` and pass it an enum to remove the view code from this type.

adrien-coye (commented):
I did a final review. All good, only one open question about navigation code in a DeeplinkParser component.

kDriveCore/Utils/DeeplinkParser.swift
adrien-coye: Abstracted navigation and unit tested in last commit.

adrien-coye (approved):
LGTM 👍

adrien-coye (approved):

adrien-coye:
Tested, all good. Merging an launching a new beta now.

adrien-coye (approved):

PR #1088, opened 2023-12-20
fix: Provide restorationIdentifier for MainTabViewController children

PhilippeWeidmann:


kDrive/UI/Controller/Files/File List/FileListViewController.swift
```diff
@@ -229,6 +229,7 @@ class FileListViewController: UIViewController, UICollectionViewDataSource, Swip
 
         navigationController?.setInfomaniakAppearanceNavigationBar()
 
+        guard viewModel != nil else { return }
```
PhilippeWeidmann, line 232: This is probably temporary. State restoration needs an overhaul

adrien-coye: Yep I agree. 
For the sake of the app's stability, I'm ok merging and using this in a beta, but IMO we need to check it more thoroughly before going live with this.

adrien-coye (approved):
Ok as a quickfix but we need to double-check this 👍

PR #1090, opened 2023-12-21
ci: New an reworked CI targets based on pure Tuist commands

adrien-coye:


adrien-coye:
Sticking to XCbuild instead here https://github.com/Infomaniak/ios-kDrive/pull/1091

PR #1093, opened 2024-01-03
fix: Prevent refetching PHAsset

PhilippeWeidmann:
Currently a new file corresponding to the PHAsset is fetched every time an upload session is restarted.
Because of how we get the PHAsset the new file has a new file and is effectively duplicated with a new name. Old files are never cleared because we only clear the new file upon upload completion.

### DIFF
--- a/kDriveCore/Data/UploadQueue/Operation/UploadOperation+PHAsset.swift
+++ b/kDriveCore/Data/UploadQueue/Operation/UploadOperation+PHAsset.swift
@@ -38,6 +38,16 @@ extension UploadOperation {
             return
         }
 
+        // Check if we are not restarting a session for an already imported asset
+        if let existingFileURL = file.pathURL,
+           fileManager.fileExists(atPath: existingFileURL.path) {
+            try transactionWithFile { file in
+                file.uploadingSession?.filePath = existingFileURL.path
+            }
+
+            return
+        }
+
         // Async load the url of the asset
         guard let url = await photoLibraryUploader.getUrl(for: asset) else {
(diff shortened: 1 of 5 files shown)


adrien-coye:
Let me test it out before I approve it.

adrien-coye:
So I did double-check, "could we simply use `NSTemporaryDirectory` ?" :
We do use for the chunks which make sense. 
Now I understand it would not be super wise to use it for files been uploaded even if it would work, re-fetching PHAsset on the fly if cleaned after a while.

kDriveCore/Data/UploadQueue/Operation/UploadOperation+PHAsset.swift
```diff
@@ -38,6 +38,16 @@ extension UploadOperation {
             return
         }
 
+        // Check if we are not restarting a session for an already imported asset
+        if let existingFileURL = file.pathURL,
+           fileManager.fileExists(atPath: existingFileURL.path) {
+            try transactionWithFile { file in
+                file.uploadingSession?.filePath = existingFileURL.path
```
adrien-coye, line 45: Ok so I tested the code, it works as-is. 
But I think it would be way cleaner to remove `file.uploadingSession.filePath`.
`filePath` It duplicates data, forces us to do an extra DB transaction, and looking at the code it looks more like dead code than anything else.

adrien-coye: So I did the refactor, feel free to merge or rollback as you wish.

adrien-coye (approved):

adrien-coye:
Got validation in DM, merging now

PR #1094, opened 2024-01-09
fix(HomeLastPicCollectionViewCell): Cancel previous request when preparing for reuse

PhilippeWeidmann:


adrien-coye (approved):
I have not ran the code, but looks good.

PR #1095, opened 2024-01-09
fix: Photo library upload will not restart if using applock

adrien-coye:


kDrive/AppDelegate+Launch.swift
```diff
@@ -41,6 +42,7 @@ extension AppDelegate {
             showLaunchFloatingPanel()
             requestAppStoreReview()
             askUserToRemovePicturesIfNecessary()
+            refreshCacheAndRestartUpload()
```
PhilippeWeidmann, line 45: When I designed `prepareRootViewController` I had a different usage in mind. For me this method should only do UI stuff

PhilippeWeidmann:
The renaming looks good. We now have a better understand of what all those methods do 👌

PhilippeWeidmann:
The renaming looks good. We now have a better understand of what all those methods do 👌

kDrive/AppDelegate+Launch.swift
adrien-coye: Changed as discussed.

PhilippeWeidmann (approved):

kDrive/AppDelegate.swift
```diff
@@ -233,9 +233,12 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, AccountManagerDeleg
 
         let currentState = RootViewControllerState.getCurrentState()
         prepareRootViewController(currentState: currentState)
-        if case .mainViewController = currentState {
+        switch currentState {
+        case .mainViewController, .appLock:
```
adrien-coye, line 237: This is not called at startup.

PhilippeWeidmann: Maybe we refactor this code so that it's called both in `applicationWillEnterForeground` and in `willFinishLaunchingWithOptions` right after `prepareRootViewController`

adrien-coye: Yep exactly what I did 👍

PhilippeWeidmann (approved):

PR #1100, opened 2024-01-19
feat: Uid file v3

PhilippeWeidmann:


kDriveCore/Data/Models/File.swift
```diff
@@ -730,13 +731,20 @@ public final class File: Object, Codable {
         return ProxyFile(driveId: driveId, id: id)
     }
 
+    public static func uid(driveId: Int, fileId: Int) -> String {
```
adrien-coye, line 734: To truly be a UID, shouldn't we use the user ID also ?

adrien-coye (commented):
All good. Just one question.

kDriveCore/Data/Models/File.swift
PhilippeWeidmann: As you guessed, `driveId` is unique so no need to go further with `userId`.

adrien-coye (approved):
Thanks for the feedback 🙌

PR #1106, opened 2024-01-29
feat: Upload v3

PhilippeWeidmann:


kDriveCore/Data/Cache/DriveFileManager.swift
```diff
@@ -49,7 +49,7 @@ public final class DriveFileManager {
         static let upload: UInt64 = 18
 
         /// Current version of the Drive Realm
-        static let drive: UInt64 = 9
+        static let drive: UInt64 = 10
```
PhilippeWeidmann, line 52: This seems wrong, I think you meant `static let upload: UInt64 = 19` ?

adrien-coye: Good catch, I tend to mix the two. Fixed in next commit.

PhilippeWeidmann:
LGTM. Waiting for other branches to merge 👍

adrien-coye (dismissed):
all good. 🚀

PR #1109, opened 2024-01-30
fix: User initiated upload is always higher than the PhotoRoll sync

adrien-coye:


kDriveCore/Data/UploadQueue/Servicies/PhotoLibraryUploader+Scan.swift
```diff
@@ -125,7 +125,7 @@ public extension PhotoLibraryUploader {
 
                 let algorithmImportVersion = currentDiffAlgorithmVersion
 
-                // New UploadFile to be uploaded
+                // New UploadFile to be uploaded. Priority is `.low`, first sync is `.medium`
```
PhilippeWeidmann, line 128:
```suggestion
                // New UploadFile to be uploaded. Priority is `.low`, first sync is `.normal`
```

PhilippeWeidmann (changes requested):

PR #1110, opened 2024-02-05
fix: Unable to do a transaction with a frozen object

adrien-coye:


kDrive/UI/Controller/Files/File List/InMemoryFileListViewModel.swift
```diff
@@ -40,10 +40,12 @@ class InMemoryFileListViewModel: FileListViewModel {
         }
 
         super.init(configuration: configuration, driveFileManager: driveFileManager, currentDirectory: currentDirectory)
+
+        guard let liveDirectory = try? currentDirectory.thaw() else { return }
```
adrien-coye, line 44: This does not fix anything. Sometimes we have a detached object and it works, sometimes we have a frozen one.
We probably should .detach all the time and it should work.

adrien-coye (commented):
Broken as is. See comment to fix for good.

adrien-coye:
Supper seeded by https://github.com/Infomaniak/ios-kDrive/pull/1118

PR #1111, opened 2024-02-06
fix: Photo sync can be gracefully terminated by the system in background

adrien-coye:
### Abstract

Photo sync was not a good citizen with the system when parsing the photo library. It now is.

I did what I could with the current state of the photo sync code that would require a good refactor.


kDriveCore/Utils/PHAsset/PHAssetIdentifier.swift
```diff
@@ -53,75 +78,110 @@ struct PHAssetIdentifier: PHAssetIdentifiable {
...
+                return nil
+            }
 
-        let group = TolerantDispatchGroup()
-        var hash: String?
+            // We build an ExpiringActivity to track system termination
+            let uid = "\(#function)-\(asset.localIdentifier)-\(UUID().uuidString)"
+            let group = TolerantDispatchGroup()
+            var error: NSError?
+            var errorPointer = NSErrorPointer(&error)
```
adrien-coye, line 90: We can have a call to discuss this if you want @PhilippeWeidmann

PhilippeWeidmann: It took me some time to understand what you are trying to achieve but in the end it seems reasonable.
I think we can do better than NSErrorPointer though

adrien-coye: I would love to do without this.
This code is relying on a blocking `TolerantDispatchGroup` I do not see how a simple error in the delagate would work around the current synchronous locked code. We should DM.

PhilippeWeidmann (approved):

PR #1113, opened 2024-02-08
feat: ShareExtension delegates uploading to Main App.

adrien-coye:
### Abstract

Since the first version of the share extension, we rely on a flag (`initiatedFromFileManager`) to distinguish uploads initiated from the share extension from the main app, since we do not want two processes to try to upload the same ressource.

For the sake of stability, only the main app uploads now, and the share extension copies files to the appGroup for upload.

kDrive now opens to start the upload after the user confirms the upload.

### DIFF
--- a/kDrive/UI/Controller/Files/Save File/SaveFileViewController.swift
+++ b/kDrive/UI/Controller/Files/Save File/SaveFileViewController.swift
@@ -479,8 +479,8 @@ extension SaveFileViewController: SelectPhotoFormatDelegate {
 
 extension SaveFileViewController: FooterButtonDelegate {
     @objc func didClickOnButton(_ sender: AnyObject) {
-        guard let selectedDriveFileManager,
-              let selectedDirectory else {
+        guard let drive = selectedDriveFileManager?.drive,
+              let directory = selectedDirectory else {
             return
         }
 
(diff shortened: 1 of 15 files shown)


kDrive/UI/Controller/Files/Save File/SaveFileViewController.swift
```diff
@@ -181,6 +181,19 @@ class SaveFileViewController: UIViewController {
         }
     }
 
+    private func dismiss(animated: Bool, clean: Bool = true, completion: (() -> Void)? = nil) {
+        Task {
+            // Cleanup file that were duplicated to appGroup on extension mode
```
adrien-coye, line 186: We could abstract where the file is stored and if it needs some form of duplication and cleanup, but I won't have the time to do it with this PR

kDriveCore/Utils/Files/FileImportHelper.swift
```diff
@@ -213,14 +231,29 @@ public final class FileImportHelper {
                 }
 
                 // Build a collection of `ImportedFile`
-                let processedFiles: [ImportedFile] = results.compactMap { taskResult in
+                let processedFiles: [ImportedFile] = try await results.concurrentCompactMap { taskResult in
```
adrien-coye, line 234: The file copy is now concurrent

adrien-coye (commented):
I did a self review to leave notes for the next reviewer.

PhilippeWeidmann (approved):

PR #1123, opened 2024-02-26
fix: ShareExtension fails to upload

adrien-coye:
Under some circumstances, it seems the upload starts from the share extension and breaks the current expectation of only the main app runs the uploads.


kDriveCore/Data/Api/DriveApiFetcher.swift
```diff
@@ -54,6 +54,7 @@ public class AuthenticatedImageRequestModifier: ImageDownloadRequestModifier {
 public class DriveApiFetcher: ApiFetcher {
     @LazyInjectService var accountManager: AccountManageable
     @LazyInjectService var tokenable: InfomaniakTokenable
+    @LazyInjectService var appContextService: AppContextServiceable
```
adrien-coye, line 57: not used

adrien-coye: fixed

valentinperignon (approved):

PR #1128, opened 2024-02-29
feat: Safer AccountManager

PhilippeWeidmann:
First PR for AccountManager improvements. The refactor of AccountManager will be splitted in multiple PRs for improved readability.

This first PRs simply uses the same thread safe types as Mail to access critical properties.

### DIFF
--- a/kDriveCore/Data/Cache/AccountManager.swift
+++ b/kDriveCore/Data/Cache/AccountManager.swift
@@ -103,12 +103,11 @@ public class AccountManager: RefreshTokenDelegate, AccountManageable {
     private static let group = "com.infomaniak.drive"
     public static let appGroup = "group." + group
     public static let accessGroup: String = AccountManager.appIdentifierPrefix + AccountManager.group
-    private let tag = "ch.infomaniak.token".data(using: .utf8)!
+
     public var currentAccount: Account!
     public var accounts = [Account]()
     public var tokens = [ApiToken]()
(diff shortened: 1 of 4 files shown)


kDriveCore/Data/Cache/AccountManager.swift
```diff
@@ -318,9 +312,8 @@ public class AccountManager: RefreshTokenDelegate, AccountManageable {
             throw DriveError.unknownToken
         }
 
-        let apiFetcher = await AccountActor.run {
```
adrien-coye, line 321: I just want to note that this was a mechanism to prevent having two `getApiFetcher` calls at once, as I understand it. 

Now, the account manager is thread safe, but we might be able to call multiple `getApiFetcher(for: account.userId, token: account.token)` in parallel.

It's probably fine, but maybe we should track the `updateUser` as a task and return if already running or something.

adrien-coye (approved):
LGTM. Just one remark, I might be overcautious, let me know.

kDriveCore/Data/Cache/AccountManager.swift
PhilippeWeidmann: This was indeed to make the access thread safe however this was kind of a hack. Now it should be safe because getApiFetcher only do calls on properties that use our threadsafe collections

adrien-coye: perfect 🤝

adrien-coye (approved):
LGTM

PR #1129, opened 2024-02-29
fix: Folder handling in file provider

adrien-coye:
### Abstract 

Large uploads in FileProvider can be flaky.

Now uploads in FileProvider will migrate to the main app when the system tries to stop the Extension.

Also made sure to compartmentalize uploads between the `App` and the `FileProvider`

### Tasks

- initiatedFromFileManager renamed to `ownedByFileProvider` for cleanliness
- Suspend FileProvider Uploads on Expiring Activity _and_ memory warnings


kDriveCore/Data/DownloadQueue/DownloadQueue.swift
```diff
@@ -22,6 +22,14 @@ import InfomaniakCore
 import InfomaniakDI
 import RealmSwift
 
+// TODO: Move to core
+extension SendableDictionary {
+    var isEmpty: Bool {
+        // swiftlint:disable empty_count
+        count > 0
```
PhilippeWeidmann, line 29: Logic is wrong here:
```suggestion
        count == 0
```

Also when moving to core don't forget to call isEmpty instead

PhilippeWeidmann (changes requested):
See my comment on single file

kDriveCore/Data/DownloadQueue/DownloadQueue.swift
adrien-coye: Good catch. Patched.
I do not intend to change core right away since as a breaking version is in the works.

PhilippeWeidmann (dismissed):

adrien-coye:
TODO: Check iOS 17 paste behaviour cc @PhilippeWeidmann

PR #1130, opened 2024-02-29
fix: Dark screen when tap under the middle TabBarButton

adrien-coye:
### Abstract

Fix the dark screen when you tap a bit under the (+) button as there is a TabBar button that links to no ViewController.

### DIFF
--- a/kDrive/UI/Controller/MainTabViewController.swift
+++ b/kDrive/UI/Controller/MainTabViewController.swift
@@ -24,7 +24,10 @@ import kDriveCore
 import kDriveResources
 import UIKit
 
-class MainTabViewController: UITabBarController, Restorable {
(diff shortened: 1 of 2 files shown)


PhilippeWeidmann (changes requested):
I think it's better to disable tabItem at controller level on init the MainTabViewController. That way we don't rely on an index.
```
let fakeViewController = UIViewController()
fakeViewController.tabBarItem.isEnabled = false
rootViewControllers.append(fakeViewController)
```

adrien-coye:
@PhilippeWeidmann Sure let me change the fix. I did it quickly last night as I saw a quikwin.

PhilippeWeidmann (approved):

PR #1091, opened 2023-12-21
ci: Rework CI based on current XCbuild command

adrien-coye:
### Abstract

Given the new test target for API tests, I'm making discrete test targets for UT / UI / API tests and we should be able to quickly see what's wrong when a test fails.

The use of dedicated CI Checks makes it easier to read and parallelise better.

### Misc

Superseding https://github.com/Infomaniak/ios-kDrive/pull/1090

~~Still in draft as needs minor rework to work with this PR that should be merged first : https://github.com/Infomaniak/ios-kDrive/pull/1092~~

PR #1132, opened 2024-03-01
ci: Upgrade script

PhilippeWeidmann:


adrien-coye:
superseeding https://github.com/Infomaniak/ios-kDrive/pull/1091

PR #1135, opened 2024-03-05
fix: Improved init

PhilippeWeidmann:
Make AccountManager more aligned with what we have in Mail


adrien-coye (commented):
I did a first review, it all looks good from a code standpoint. I'll do some quick Q/A this afternoon.

adrien-coye (approved):
Did a non regression test on the iPhone 8, it works well. All good to me.

PR #1144, opened 2024-03-11
fix: Fix fetch request to clean uploaded pictures

adrien-coye:


kDriveCore/Data/Cache/BackgroundRealm.swift
```diff
@@ -81,9 +81,12 @@ public class BackgroundRealm {
     }
 
     public func execute(_ block: (Realm) -> Void) {
+        let activity = ExpiringActivity()
```
adrien-coye, line 84: I added this quick win to make sure it won't create 'deadlock crashes' since more work is performed asynchronously.

adrien-coye (commented):
Did a self review.

kDriveCore/Data/Cache/BackgroundRealm.swift
PhilippeWeidmann: We've had this mecanism for a long time now on Mail. I wonder why we didn't implement it earlier in kDrive

PR #1153, opened 2024-03-15
fix: EmptyView not showing

PhilippeWeidmann:
Empty view fix


adrien-coye (commented):
OK regarding the code. 

We discussed that I still have an empty view issue, after creating a new folder.

adrien-coye (approved):
I still got some strange behaviour in offline mode. 
Yet all good for now, this calls for a deeper refactor.

PR #1161, opened 2024-03-27
fix: Avoid crash offline files & co + Avoid crash Trash

PhilippeWeidmann:


adrien-coye (approved):
Easy to review commit by commit.

PR #1163, opened 2024-04-04
feat: Null avatar added

Ambrdctr:


adrien-coye (approved):
LGTM 👍 
- discussed: Tested with Null avatar.

PR #1165, opened 2024-04-05
chore: Clean FileProviderExtensionAdditionalState.swift

adrien-coye:


adrien-coye:
Done on master instead of APIV3

PR #1167, opened 2024-04-05
feat: Safe realm access

adrien-coye:
### Abstract

Realm issues encountered came from the access pattern to it. 
We recently started to explicit frozen and detached in the name of realm objects.
This PR goes further by introducing a `transaction style API` to make sure we distinguish reading from writing transactions. We no longer pass around a random realm object in a random state.

This API also allows us to ask the system to not suddenly terminate the app while we are performing a database transaction, which crashes in the current version of the app.

### Linked PRs 

Sub PR for another object refactored similarly since this PR is already a thousand line ish.
https://github.com/Infomaniak/ios-kDrive/pull/1168

Core PR to create a CoreDB library with the new transaction style protocols.
https://github.com/Infomaniak/ios-core/pull/121/


kDriveCore/Data/Cache/DriveFileManager+Realm.swift
```diff
@@ -0,0 +1,106 @@
...
+
+        let realm = getRealm()
+        realmClosure(realm)
+    }
+
+    // TODO: This should be private
+    /// Get an up to date realm for current DriveFileManager
+    ///
+    /// This is _not_ protected from sudden termination, prefer `transaction(withRealm:)` method
+    /* private */ func getRealm() -> Realm {
```
adrien-coye, line 83: There is some work to refactor the methods that take an optional realm as a parameter, but first I focus on preventing direct access to `getRealm()`.

Using a transaction pattern, I can wrap ExpiringActivity calls to harden the transactions. That is the first stept.

kDriveCore/Data/Cache/DriveFileManager+Realm.swift
```diff
@@ -0,0 +1,130 @@
...
+            return DriveFileManager.constants.rootDocumentsURL.appendingPathComponent("\(drive.userId)-\(drive.id)-fp.realm")
+        }
+    }
+}
+
+public extension DriveFileManager {
+    /// Common way to do a __write__ transaction with the current Realm of the DriveFileManager
+    ///
+    /// Protected from sudden termination
+    func writeTransaction(withRealm realmClosure: (Realm) throws -> Void) throws {
```
adrien-coye, line 50: That's the "write transaction" I came up with. I might be able to make the API a bit more flexible. Still thinking about it.

kDriveCore/Data/Cache/DriveFileManager+Realm.swift
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+    }
+
+    /// Common way to do a read transaction, to fetch one entity, with the current Realm of the DriveFileManager
+    ///
+    /// Protected from sudden termination
+    ///
+    /// NSException thrown if mutating realm elements
+    func fetchObject<Element: Object>(ofType type: Element.Type,
```
adrien-coye, line 70: That's a straight forward fetch request for a single entity, this one works well as is, and I find it elegant and easy to use.

kDriveCore/Data/Cache/DriveFileManager+Realm.swift
```diff
@@ -0,0 +1,130 @@
...
+            return try realmClosure(realm)
+        }
+    }
+
+    /// Common way to do a read transaction, to fetch a collection, with the current Realm of the DriveFileManager
+    ///
+    /// Protected from sudden termination
+    ///
+    /// NSException thrown if mutating realm elements
+    func fetchResults<Element: RealmFetchable>(ofType type: Element.Type,
```
adrien-coye, line 89: same thing with a collection.

adrien-coye (commented):
Left some notes for reviewers.

kDriveCore/Data/Cache/DriveFileManager+Realm.swift
adrien-coye: now private

Project.swift
```diff
@@ -24,7 +24,10 @@ let project = Project(name: "kDrive",
                       packages: [
                           .package(url: "https://github.com/apple/swift-algorithms", .upToNextMajor(from: "1.2.0")),
                           .package(url: "https://github.com/Alamofire/Alamofire", .upToNextMajor(from: "5.2.2")),
-                          .package(url: "https://github.com/Infomaniak/ios-core", .upToNextMajor(from: "9.0.0")),
+                          .package(
```
adrien-coye, line 27: TODO: Publish Core

kDrive/UI/Controller/Files/Categories/ManageCategoriesViewController.swift
```diff
@@ -252,8 +252,14 @@ final class ManageCategoriesViewController: UITableViewController {
             return
         }
         self.driveFileManager = driveFileManager
-        let realm = driveFileManager.getRealm()
-        files = filesId.compactMap { driveFileManager.getCachedFile(id: $0, using: realm) }
+
+        let matchedFiles = driveFileManager.fetchResults(ofType: File.self) { realm in
+            realm.objects(File.self)
+                .filter("id IN %@", filesId)
```
adrien-coye, line 258: I migrated to a NSPredicate to be able to use a read transaction

kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
```diff
@@ -609,8 +609,13 @@ class PreviewViewController: UIViewController, PreviewContentCellDelegate {
         }
         self.driveFileManager = driveFileManager
         let previewFileIds = coder.decodeObject(forKey: "Files") as? [Int] ?? []
-        let realm = driveFileManager.getRealm()
-        previewFiles = previewFileIds.compactMap { driveFileManager.getCachedFile(id: $0, using: realm) }
+
+        let matchedFiles = driveFileManager.fetchResults(ofType: File.self) { realm in
+            realm.objects(File.self)
+                .filter("id IN %@", previewFileIds)
```
adrien-coye, line 615: I migrated to a NSPredicate to be able to use a read transaction

kDrive/UI/Controller/Files/Categories/ManageCategoriesViewController.swift
PhilippeWeidmann: Maybe we want to use the strongly typed         
```suggestion
                .where {
                    $0.id.in(filesId)
                }
```

PhilippeWeidmann:
In general, what do you think about using more generally the realm's NSPredicate generator, it does not rely on strings and is strongly typed.
For example, to filter we use `Results<Model>.where { $0.id == 1 }`

kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+AvailableOffline.swift
```diff
@@ -21,13 +21,14 @@ import RealmSwift
...
-            .sorted(by: [sortType.value.sortDescriptor]).freeze()
+        let frozenFiles = fetchResults(ofType: File.self) { realm in
+            return realm.objects(File.self)
+                .filter("isAvailableOffline = true")
+                .sorted(by: [sortType.value.sortDescriptor])
+                .freeze()
+        }
 
-        return offlineFiles.map { $0.freeze() }
+        return frozenFiles.compactMap { !$0.isInvalidated ? $0 : nil }
```
PhilippeWeidmann, line 31: We shouldn't need to check for isInvalidated because we are using the objects we've just fetched from Realm

PhilippeWeidmann: (Maybe there are also other places in the modifications where you can remove it)

kDriveCore/Data/Database/RealmAccessor.swift
```diff
@@ -0,0 +1,58 @@
+//
```
PhilippeWeidmann, line 1:
```suggestion

```

adrien-coye:
> In general, what do you think about using more generally the realm's NSPredicate generator, it does not rely on strings and is strongly typed.
> For example, to filter we use `Results<Model>.where { $0.id == 1 }`

Depends on how committed to abstracting the DB we are. This is a nice API to use I admit.
I do see the value in using type validation / strong types to generate a predicate, but this is a feature from Realm, furthering our vendor locking.

kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+AvailableOffline.swift
adrien-coye: I think it was added as I saw a crash, as if a deletion is probably performed on another thread it could be invalidated here, so I'm on the fence about this one.

kDrive/UI/Controller/Files/Categories/ManageCategoriesViewController.swift
adrien-coye: Well I'm not sure, it depends what we want to do abstraction wise 

https://github.com/Infomaniak/ios-kDrive/pull/1167#issuecomment-2058623898

kDrive/UI/Controller/Files/Categories/ManageCategoriesViewController.swift
```diff
@@ -252,8 +252,13 @@ final class ManageCategoriesViewController: UITableViewController {
...
         }
         self.driveFileManager = driveFileManager
-        let realm = driveFileManager.getRealm()
-        files = filesId.compactMap { driveFileManager.getCachedFile(id: $0, using: realm) }
+
+        let matchedFiles = driveFileManager.fetchResults(ofType: File.self) { faultedCollection in
+            faultedCollection.filter("id IN %@", filesId)
+        }
+
+        files = matchedFiles.compactMap { !$0.isInvalidated ? $0 : nil }
```
PhilippeWeidmann, line 260: Still not convince by this, from my point of view objects cannot be invalidated right after being pulled out of a realm. I will do some tests on my side.

kDrive/UI/Controller/Files/RootMenuViewController.swift
```diff
@@ -122,8 +122,13 @@ class RootMenuViewController: CustomLargeTitleCollectionViewController, SelectSw
         configureDataSource()
 
         let rootFileUid = File.uid(driveId: driveFileManager.drive.id, fileId: DriveFileManager.constants.rootID)
-        let rootChildren = driveFileManager.getRealm().object(ofType: File.self, forPrimaryKey: rootFileUid)?.children
-        rootChildrenObservationToken = rootChildren?.observe { [weak self] changes in
+        guard let root = driveFileManager.fetchObject(ofType: File.self, forPrimaryKey: rootFileUid),
+              !root.isInvalidated else {
```
PhilippeWeidmann, line 126: Same here (https://github.com/Infomaniak/ios-kDrive/pull/1167/files#r1584310026)

kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+AvailableOffline.swift
PhilippeWeidmann: Everything happen serially within those calls so the objects cannot be invalid here

kDriveCore/Data/Api/Endpoint.swift
```diff
@@ -133,10 +133,34 @@ public struct ProxyFile: AbstractFile, Sendable {
         self.id = id
     }
 
-    func resolve(using realm: Realm) throws -> File {
+    /// Resolve an abstract file within a `DriveFileManager`.
+    func resolve(within driveFileManager: DriveFileManager) throws -> File {
+        let file = driveFileManager.fetchObject(ofType: File.self) { realm in
+            fetch(using: realm)
+        }
```
PhilippeWeidmann, line 140: I have a crash here, this call seems weird.
Using this seems better:
```suggestion
       let file = driveFileManager.fetchObject(ofType: File.self, forPrimaryKey: uid)
```

kDriveCore/Data/Api/Endpoint.swift
```diff
@@ -133,10 +133,34 @@ public struct ProxyFile: AbstractFile, Sendable {
...
+
+        guard let file, !file.isInvalidated else {
```
PhilippeWeidmann, line 142: More invalidated check here

kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+Query.swift
```diff
@@ -0,0 +1,126 @@
...
+        let uid = File.uid(driveId: drive.id, fileId: id)
+        guard let file = fetchObject(ofType: File.self, forPrimaryKey: uid), !file.isInvalidated else {
+            return nil
+        }
+        return freeze ? file.freeze() : file
+    }
+
+    func getCachedFile(id: Int, freeze: Bool = true, using realm: Realm) -> File? {
+        let uid = File.uid(driveId: drive.id, fileId: id)
+        guard let file = realm.object(ofType: File.self, forPrimaryKey: uid), !file.isInvalidated else {
```
PhilippeWeidmann, line 66: Here it's okay to check if file is invalidated because realm could be live and could have changed

kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+Query.swift
```diff
@@ -0,0 +1,126 @@
...
+        guard let file, !file.isInvalidated else {
+            return nil
+        }
+
+        return file
+    }
+
+    func getCachedFile(id: Int, freeze: Bool = true) -> File? {
+        let uid = File.uid(driveId: drive.id, fileId: id)
+        guard let file = fetchObject(ofType: File.self, forPrimaryKey: uid), !file.isInvalidated else {
```
PhilippeWeidmann, line 58: More invalidated check here

kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+Query.swift
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    func getCachedMyFilesRoot() -> File? {
+        let file = try? fetchObject(ofType: File.self) { faultedCollection in
+            faultedCollection.filter("rawVisibility == %@", FileVisibility.isPrivateSpace.rawValue)
+                .first?
+                .freeze()
+        }
+
+        guard let file, !file.isInvalidated else {
```
PhilippeWeidmann, line 49: More invalidated check here

PhilippeWeidmann (changes requested):
There are too many unnecessary checks for invalidation.
An object can only become invalidated in a live realm. 

If a realm is frozen it's okay to have invalidated objects.
If we try to resolve a frozen object in a new live fresh realm the object will simply not be returned.

Here is a small example to show how realm works:
```swift
        let testFile = File(id: -100, name: "Some test file", driveId: 1)
        try? writeTransaction { writableRealm in
            writableRealm.add(testFile)
        }
        
        let testFrozenFile = testFile.freezeIfNeeded()

        try? writeTransaction { writableRealm in
            writableRealm.delete(testFile)
        }
        
        print("[Test]", testFrozenFile.isInvalidated, testFrozenFile.uid)
        // [Test] false -100_1
```

adrien-coye:
@PhilippeWeidmann Sure we can surely do without the `.isInvalidated` checks since we _now_ are way more rigorous with our use of live realm objects 👍 

I'm pretty sure it was very useful in previous versions of the app, and it was made as a fix to numerous Sentry bug reports.

kDriveCore/Data/Api/Endpoint.swift
adrien-coye: Yeah but this one is live so I keep it right ?

kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+Query.swift
adrien-coye: okay

kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+Query.swift
adrien-coye: no longer a realm here, so ok.

kDrive/UI/Controller/Files/Categories/ManageCategoriesViewController.swift
adrien-coye: Yes. Since this comes from a throw away realm / Transactionable, it is fine without it.

kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
```diff
@@ -609,8 +609,12 @@ class PreviewViewController: UIViewController, PreviewContentCellDelegate {
...
         self.driveFileManager = driveFileManager
         let previewFileIds = coder.decodeObject(forKey: "Files") as? [Int] ?? []
-        let realm = driveFileManager.getRealm()
-        previewFiles = previewFileIds.compactMap { driveFileManager.getCachedFile(id: $0, using: realm) }
+
+        let matchedFiles = driveFileManager.fetchResults(ofType: File.self) { faultedCollection in
+            faultedCollection.filter("id IN %@", previewFileIds)
+        }
+
+        previewFiles = matchedFiles.compactMap { !$0.isInvalidated ? $0 : nil }
```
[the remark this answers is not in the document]
adrien-coye: Double checked.
Since this comes from a throw away realm / Transactionable, at this exact moment it cannot be invalidated.

kDrive/UI/Controller/Files/RootMenuViewController.swift
adrien-coye: Double checked.
Since this comes from a throw away realm / Transactionable, it is fine without it.

kDriveCore/Data/Api/Endpoint.swift
adrien-coye: Good catch, indeed it is better.

PhilippeWeidmann (approved):

adrien-coye:
Only cleaned some Sonar issues. Squash merging now.

PR #1173, opened 2024-04-23
fix: Fix import folder taking a lot of space

adrien-coye:
### Abstract

A copy of a file was kept when an upload was cancelled. This is no longer the case.

### DIFF
--- a/kDriveCore/Data/DownloadQueue/DownloadQueue.swift
+++ b/kDriveCore/Data/DownloadQueue/DownloadQueue.swift
@@ -212,10 +212,6 @@ public final class DownloadQueue: ParallelismHeuristicDelegate {
         operationQueue.cancelAllOperations()
     }
 
-    public func cancelRunningOperations() {
-        operationQueue.operations.filter(\.isExecuting).forEach { $0.cancel() }
-    }
-
     /// Check if a file is been uploaded
     ///
     /// Thread safe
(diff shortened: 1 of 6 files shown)


kDriveCore/Data/DownloadQueue/DownloadQueue.swift
```diff
@@ -212,10 +212,6 @@ public final class DownloadQueue: ParallelismHeuristicDelegate {
         operationQueue.cancelAllOperations()
     }
 
-    public func cancelRunningOperations() {
```
adrien-coye, line 215: This was dead code

kDriveCore/Data/UploadQueue/Queue/UploadQueue+Queue.swift
```diff
@@ -22,60 +22,6 @@ import Foundation
 import InfomaniakCore
 import RealmSwift
 
-public protocol UploadQueueable {
```
adrien-coye, line 25: I found navigation hard, so I moved the protocol to a dedicated file

kDriveFileProvider/FileProviderExtension.swift
```diff
@@ -291,7 +291,7 @@ final class FileProviderExtension: NSFileProviderExtension {
             url: item.storageUrl,
             name: item.filename,
             conflictOption: .version,
-            shouldRemoveAfterUpload: false
+            shouldRemoveAfterUpload: true
```
adrien-coye, line 294: As discussed, upload from the FileProvider should not keep a file copy.

PR #1176, opened 2024-04-30
fix: Clean ophan files on app exit on APIV3

adrien-coye:
Making sure files in the import folder are used by an upload in progress tracked in DB, else the file is orphan and meets the void.

Note: The upload queue must be stopped to perform this work, therefore processing it when closing the app is preferable for now.

Will be done on top of master instead. Possible in the end, unlike originally thought. [PR](https://github.com/Infomaniak/ios-kDrive/pull/1177)


PhilippeWeidmann (approved):
LGTM. One small remark, this only runs when the app is killed. Maybe we want to run it more often ?

adrien-coye:
Will be done on top of master instead. Possible in the end, unlike originally thought. https://github.com/Infomaniak/ios-kDrive/pull/1177

PR #1181, opened 2024-05-16
feat: Switching drive stays in the current tab

adrien-coye:
Switching drives (within one account) does not restore the correct tab on APIV3 branch.
This takes care of that.

A further PR will refactor the app restoration to make good use of Scenes.


kDrive/Utils/AppFactoryService.swift
```diff
@@ -20,19 +20,108 @@ import Foundation
...
+    public func shouldRestoreApplicationState(coder: NSCoder) -> Bool {
+        let encodedVersion = coder.decodeInteger(forKey: Self.appStateVersionKey)
+        let shouldRestoreApplicationState = Self.currentStateVersion == encodedVersion &&
+            !(UserDefaults.shared.legacyIsFirstLaunch || accountManager.accounts.isEmpty)
+        Log.appDelegate("shouldRestoreApplicationState:\(shouldRestoreApplicationState)")
+        return shouldRestoreApplicationState
+    }
+
+    public func reloadAppUI(for drive: Drive) {
+        @InjectService var accountManager: AccountManageable
```
adrien-coye, line 85: duplicated code

kDrive/Utils/AppFactoryService.swift
```diff
@@ -20,19 +20,108 @@ import Foundation
...
 
         // setup DI ASAP
         FactoryService.setupDependencyInjection(other: extraDependencies)
     }
 }
+
+// TODO: Refactor with Scenes / NSUserActivity
+#if !ISEXTENSION
+
+public final class AppRestorationService {
```
adrien-coye, line 50: move this code outside this file, probably requires some project changes but it will be cleaner.

adrien-coye (commented):
Self assesment.

kDrive/Utils/AppFactoryService.swift
adrien-coye: reworked a bit to tidy this up

kDrive/AppRestorationService.swift
```diff
@@ -0,0 +1,79 @@
...
+        coder.encode(Self.currentStateVersion, forKey: Self.appStateVersionKey)
+        return true
+    }
+
+    public func shouldRestoreApplicationState(coder: NSCoder) -> Bool {
+        let encodedVersion = coder.decodeInteger(forKey: Self.appStateVersionKey)
+        let shouldRestoreApplicationState = Self.currentStateVersion == encodedVersion &&
+            !(UserDefaults.shared.legacyIsFirstLaunch || accountManager.accounts.isEmpty)
+        Log.appDelegate("shouldRestoreApplicationState:\(shouldRestoreApplicationState)")
+        return shouldRestoreApplicationState
```
PhilippeWeidmann, line 56: Maybe we should still return false for now because state restoration was causing issues with viewmodel init

kDrive/AppRestorationService.swift
```diff
@@ -0,0 +1,79 @@
...
+        // This should be migrated to NSUserActivity at some point
+        let lastSelectedTab = UserDefaults.shared.lastSelectedTab
+        let newMainTabViewController = MainTabViewController(
+            driveFileManager: currentDriveFileManager,
+            selectedIndex: lastSelectedTab
+        )
+        (UIApplication.shared.delegate as? AppDelegate)?.setRootViewController(newMainTabViewController)
+
+        // cleanup
+        UserDefaults.shared.lastSelectedTab = nil
```
PhilippeWeidmann, line 77: We should maybe do it on switch account because for now `lastSelectedTab` stays nil if we switch drive before `didSelect viewController` is called

PhilippeWeidmann (approved):

PR #1188, opened 2024-05-24
feat(LockedAppViewController): Unlock button should be loading while evaluating policy

valentinperignon:
https://github.com/Infomaniak/ios-kDrive/assets/25110022/f8bd59c9-80f4-4fcd-a447-7126323d8d61


adrien-coye (approved):
Looks good to me, I'll play a bit with unlocking when I'll publish the next beta.

PR #1196, opened 2024-06-07
fix: Navigate button fixed for APIV3 view changes

adrien-coye:
"navigate" action on a snackbar, after an upload is finished was broken on APIV3 branch.
This fixes that.

### DIFF
--- a/kDrive/AppDelegate.swift
+++ b/kDrive/AppDelegate.swift
@@ -387,32 +387,52 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, AccountManagerDeleg
                 return
             }
 
-            if !file.isRoot && viewController.viewModel.currentDirectory.id != file.id {
-                // Pop to root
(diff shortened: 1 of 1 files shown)


kDrive/AppDelegate.swift
```diff
@@ -387,32 +387,52 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, AccountManagerDeleg
...
+                  viewController.viewModel.currentDirectory.id != file.id else {
+                Log.appDelegate("Already presenting the correct screen")
+                return
+            }
+
+            // Pop to root
+            navController.popToRootViewController(animated: false)
+
+            // Present file
+            guard let fileListViewController = navController.topViewController as? RootMenuViewController else {
```
adrien-coye, line 400: This line is the fix. The rest is refactor of the code to use gards.

valentinperignon (approved):

PR #1197, opened 2024-06-11
fix: Ensure we enter before trying to retrieve the resource

PhilippeWeidmann:
Sometimes the upload queue is stuck when calling `bestResourceSHA256` or `baseImageSHA256` this happens if pictures are stored on iCloud. 
From what I've observed it is because `PHAssetResourceManager.default().requestData` or `asset.requestContentEditingInput` returns directly instead of switching threads. We end up calling group.leave before group.enter and so we end up waiting indefinitely on group.wait. 
This would result in a soft lock blocking all cooperative thread pools.


adrien-coye (approved):
Thanks for catching this 👍

PR #1200, opened 2024-06-13
feat: Offline files

PhilippeWeidmann:
Waiting for prod endpoints


adrien-coye (commented):
Only reviewed the code. Looks good.

kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+AvailableOffline.swift
```diff
@@ -35,81 +35,44 @@ public extension DriveFileManager {
...
-
-            for deletedFileUid in updatedFileResult.deletedFileUids {
-                deleteOfflineFile(uid: deletedFileUid, writableRealm: writableRealm)
+            var updatedFiles = [File]()
+            try database.writeTransaction { writableRealm in
+                for activity in batchActivities {
+                    if let file = activity.file,
+                       [FileActivityType.fileUpdate, FileActivityType.fileRename].contains(activity.lastAction) {
+                        updateFile(updatedFile: file, lastActionAt: activity.lastActionAt, writableRealm: writableRealm)
+                        updatedFiles.append(file)
```
adrien-coye, line 49: Once the files are attached to a realm, can we freeze them, or am I being overcautious ?

adrien-coye (commented):
I did another review of the code and did some QA on iPhone 8.
I'll re-review later with pull to refresh 👍

kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+AvailableOffline.swift
PhilippeWeidmann: Good catch. We could be on other thread for each batch

PhilippeWeidmann: It's okay to not freeze it since we only use it on the same thread ln:60. Between each array updatedFiles is recreated so it's safe even if we switch threads between await.

PR #1208, opened 2024-07-04
fix: Correctly load root on switch

PhilippeWeidmann:
Root is loaded for the current drive and the current account on update but not once we switch drive. 
Also the root was loaded twice in `initRoot` by 2 different calls.

This PR ensures we load in the background the root while we switch.
It also removes the second call.


adrien-coye (approved):
All good. I'll test the code with the private beta.

PR #1209, opened 2024-07-04
fix: Use revisedAt instead of lastModifiedAt where possible

PhilippeWeidmann:
Now that we have access to `revisedAt` which gives a date for the modification of the file we can use it to compare with the local file date.


adrien-coye (approved):
All good. Thanks

PR #1211, opened 2024-07-04
fix: Add missing fileSupportedBy for activities

PhilippeWeidmann:
This reenables thumbnails in home activities


adrien-coye (approved):
👍

PR #1217, opened 2024-07-05
fix: Extension main activity missing

adrien-coye:
Share extension is crashing since the SceneDelegate migration. This fixes that.


adrien-coye (commented):
Did a test on device, merging for next beta.

PR #1220, opened 2024-07-08
fix: Use correct endpoints

PhilippeWeidmann:
Read commit by commit:
- (417cc90fabb154f77d13f951ed00a9a89d76d902) Trash -> restore / delete permanently use V2
- (12387860dc3153893c21d3d95fdb98a4024f603a) Activities for file details -> use cursor
- (5391da64ea1b9e13bcd68d00e0a6447f7437a681) Remove dead code
- (f6df90dfabfca9a40956c36a068b180da95a2964) Remove default argument for searchFiles

### DIFF
--- a/kDriveAPITests/kDriveCore/DriveApiTests.swift
+++ b/kDriveAPITests/kDriveCore/DriveApiTests.swift
@@ -608,26 +608,6 @@ final class DriveApiTests: XCTestCase {
         tearDownTest(directory: testDirectory)
     }
 
-    func testGetFilesActivities() async throws {
-        let (testDirectory, file) = try await initOfficeFile(testName: "Get files activities")
-        let secondFile = try await currentApiFetcher.createFile(
-            in: testDirectory,
-            name: "Get files activities-\(Date())",
-            type: "docx"
-        ).proxify()
-        let activities = try await currentApiFetcher.filesActivities(
-            drive: proxyDrive,
-            files: [file, secondFile],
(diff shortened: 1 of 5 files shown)


kDriveAPITests/kDriveCore/DriveApiTests.swift
```diff
@@ -597,34 +597,14 @@ final class DriveApiTests: XCTestCase {
...
-            drive: proxyDrive,
-            files: [file, secondFile],
-            from: Date(timeIntervalSince1970: 0)
-        ).validApiResponse.data
-        XCTAssertEqual(activities.count, 2, "Array should contain two activities")
-        for activity in activities {
-            XCTAssertTrue(activity.result, TestsMessages.shouldReturnTrue)
-            XCTAssertNil(activity.message, TestsMessages.noError)
-        }
+        _ = try await currentApiFetcher.fileActivities(file: file, from: earlyDate, cursor: nil)
```
adrien-coye, line 607: No assertions on the result of this call?

adrien-coye (commented):
Thanks for fixing some routes, I only got one tiny comment.

kDriveAPITests/kDriveCore/DriveApiTests.swift
PhilippeWeidmann: I think the goal of this test was juste to check if we could do the API call and decode the result correctly

adrien-coye (approved):
All set.

PR #1221, opened 2024-07-09
fix: Correctly name private root

PhilippeWeidmann:
Private space root is named differently based on drivePack


adrien-coye (approved):
I did not run the code, but it looks OK to me.

PR #1225, opened 2024-07-11
fix: Keep sortedName as lowercased

Ambrdctr:
Realm can't sort files correctly: [issue](https://github.com/realm/realm-swift/issues/2970).
Keep sortedName as lowercased to sort files alphabetically (and case insensitive).


adrien-coye (commented):
I'm OK with this code. I'm double-checking that this is what is expected for iOS and Android. Edit: Done

adrien-coye (approved):
All good 🙌

PR #1234, opened 2024-07-25
fix: Extend switch drive tappable area

Lnamw:
### Switch drive from the MenuViewController

**Before**
Tap only available on the button

**After**
Click anywhere on the stackView (icon, label and button)

Simulator Screenshot - iPhone 15 - 2024-07-25 at 09 55 00


adrien-coye (approved):
I played a bit with it, it works well. All good for me 👍

PR #1239, opened 2024-07-30
fix: Attempt to fix what looks like a realm race condition by doing a refactor

adrien-coye:
refactor: Removed some legacy GCD code


PhilippeWeidmann (approved):
Changes look okay. Cool to say good bye to this last bits 🤞

PR #1241, opened 2024-07-30
refactor: Programmatic init FilelistViewController

PhilippeWeidmann:


kDrive/UI/View/Header view/FilesHeaderView.swift
```diff
@@ -73,6 +74,10 @@ class FilesHeaderView: UICollectionReusableView {
         )
         let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(didTapOnCard))
         uploadCardView.addGestureRecognizer(tapGestureRecognizer)
+
+        selectView = SelectView.instantiate()
```
adrien-coye, line 78: So I guess your intent is to keep the `Class.instantiate()` pattern only for xib related views ?

adrien-coye (commented):
I read the code, all good. I'll play with it tomorrow.

kDrive/UI/View/Header view/FilesHeaderView.swift
PhilippeWeidmann: Yes for now I didn't to touch the XIBs 👍

adrien-coye (approved):
I did play a bit with it. All good.

PR #1242, opened 2024-07-31
fix: Fetch activities using listing

PhilippeWeidmann:
In API V2 after an action, for example an undo. We were fetching activities to update realm.

With V3 there is no notion of activities we only continue listing files based on previous cursor.


adrien-coye (approved):
All good 🙌

PR #1247, opened 2024-08-06
fix(DriveFileManager): Func deleteOrphan crashes realm

adrien-coye:


adrien-coye:
Predicate is actually correct

PR #1250, opened 2024-08-08
fix: Audio preview

adrien-coye:
### Abstract

The audio preview cell has a flaky behaviour. 

This PR aims at abstracting the "Player" away from the Cell code.
Then I fixed the issue related to the slider.

Note: Moved some of `MatomoUtils` to core in order to use it from the "Single Track Player"

Tested on device and mac.


kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
```diff
@@ -213,13 +213,15 @@ class PreviewViewController: UIViewController, PreviewContentCellDelegate, Scene
...
     func observeFileUpdated() {
         driveFileManager?.observeFileUpdated(self, fileId: nil) { [weak self] file in
-            if self?.currentFile.id == file.id {
-                self?.currentFile = file
-                Task { @MainActor in
-                    guard let self else { return }
-                    self.collectionView.endEditing(true)
-                    self.collectionView.reloadItems(at: [self.currentIndex])
-                }
+            guard let self = self,
```
PhilippeWeidmann, line 216: This could be simplified to `guard let self` ?

kDrive/UI/Controller/Menu/AppLockSettingsViewController.swift
```diff
@@ -18,6 +18,7 @@
 
 import kDriveResources
 import LocalAuthentication
+import kDriveCore
```
PhilippeWeidmann, line 21: Why do we now need core ?

kDrive/Utils/MatomoUtils+UI.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    static func trackShareLinkSettings(protectWithPassword: Bool, downloadFromLink: Bool, expirationDateLink: Bool) {
+        MatomoUtils.track(eventWithCategory: .shareAndRights, name: "protectWithPassword", value: protectWithPassword)
+        MatomoUtils.track(eventWithCategory: .shareAndRights, name: "downloadFromLink", value: downloadFromLink)
+        MatomoUtils.track(eventWithCategory: .shareAndRights, name: "expirationDateLink", value: expirationDateLink)
+    }
+
+    // MARK: - File action
+
+    #if !ISEXTENSION
```
PhilippeWeidmann, line 46: Do we still need the ifdef ? I would be great to remove them where we can as Xcode seems a bit lost with the code.

kDrive/UI/Controller/Menu/AppLockSettingsViewController.swift
adrien-coye: `MatomoUtils` in now in core. (as needed by the player)

kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
adrien-coye: Yup, I wonder why the linter does not autocorrect this.

kDrive/Utils/MatomoUtils+UI.swift
adrien-coye: If I remove the ifdef, the extensions are not seeing `FloatingPanelAction`. I can dig deeper or stop there.

kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
PhilippeWeidmann: Yup, I will check if it's a rule we need to enable

kDrive/Utils/MatomoUtils+UI.swift
PhilippeWeidmann: If you need to do a bigger refactor it's not worth it for now

PhilippeWeidmann (approved):

PR #1261, opened 2024-08-23
fix: Prevent adding parent to itself when handling actions

PhilippeWeidmann:


adrien-coye (approved):
Nice, let's go for a release this weekend with it.

PR #1265, opened 2024-08-27
fix(Files.app): Drives are hidden on soft disable from the app

adrien-coye:
Soft disable of Files.App from kDrive should hide all accounts.

### DIFF
--- a/kDrive/UI/Controller/Menu/SecurityTableViewController.swift
+++ b/kDrive/UI/Controller/Menu/SecurityTableViewController.swift
@@ -74,7 +74,25 @@ class SecurityTableViewController: BaseGroupedTableViewController {
             cell.titleLabel.text = KDriveResourcesStrings.Localizable.fileProviderExtensionTitle
             cell.detailsLabel.text = KDriveResourcesStrings.Localizable.fileProviderExtensionDescription
             cell.switchHandler = { sender in
-                UserDefaults.shared.isFileProviderExtensionEnabled = sender.isOn
+                let toggleStatus = sender.isOn
+                cell.valueSwitch.isEnabled = false
+                Task {
+                    UserDefaults.shared.isFileProviderExtensionEnabled = toggleStatus
+
+                    if toggleStatus {
+                        // Force refresh accounts, to regenerate Files.app accounts
+                        @InjectService var accountManager: AccountManageable
+                        let accounts = accountManager.accounts.values
+                        for account in accounts {
+                            _ = try? await accountManager.updateUser(for: account, registerToken: true)
+                        }
+                    } else {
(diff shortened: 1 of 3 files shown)


kDrive/UI/Controller/Menu/SecurityTableViewController.swift
```diff
@@ -74,7 +74,25 @@ class SecurityTableViewController: BaseGroupedTableViewController {
             cell.titleLabel.text = KDriveResourcesStrings.Localizable.fileProviderExtensionTitle
             cell.detailsLabel.text = KDriveResourcesStrings.Localizable.fileProviderExtensionDescription
             cell.switchHandler = { sender in
-                UserDefaults.shared.isFileProviderExtensionEnabled = sender.isOn
+                let toggleStatus = sender.isOn
+                cell.valueSwitch.isEnabled = false
```
adrien-coye, line 78: Not in love with doing this way, but it does the job. On re-enable it is necessary given I do a some network calls (if many accounts)

kDrive/UI/Controller/Menu/SecurityTableViewController.swift
```diff
@@ -74,7 +74,25 @@ class SecurityTableViewController: BaseGroupedTableViewController {
...
+                Task {
+                    UserDefaults.shared.isFileProviderExtensionEnabled = toggleStatus
+
+                    if toggleStatus {
+                        // Force refresh accounts, to regenerate Files.app accounts
+                        @InjectService var accountManager: AccountManageable
+                        let accounts = accountManager.accounts.values
+                        for account in accounts {
+                            _ = try? await accountManager.updateUser(for: account, registerToken: true)
```
adrien-coye, line 87: There is no `refreshAll()` guess I could make one. Or is there a better way to do it @PhilippeWeidmann ?

PhilippeWeidmann: Didn't find a refreshAll. Maybe in Mail.

kDrive/UI/Controller/Menu/SecurityTableViewController.swift
PhilippeWeidmann: Do we really need to do a network call ? 
If we are in this settings, Accounts / Drives should be up to date

We could simply take the drives we have for each user and add the domains.

(We can keep the Task because this process is still async but not network dependant)

PhilippeWeidmann (approved):

kDrive/UI/Controller/Menu/SecurityTableViewController.swift
adrien-coye: replied to this

PR #1267, opened 2024-08-28
feat: Add RequestContextIdAdaptor to ApiFetcher

PhilippeWeidmann:
Already added since a long time on Mail. I think we forgot to add it on Drive.


adrien-coye (approved):
👍

PR #1268, opened 2024-08-29
fix: Pass correct activities for FileDetailsView

PhilippeWeidmann:
This PR addresses two issues in the FileActivityDetails:
- We only want to display a subset of the activities in the file details view. 
- Some of the activities we asked for were also invalid.


adrien-coye (approved):
THX 1138

PR #1273, opened 2024-09-04
fix: FileProvider upload already existing files + Correctly report missing thumbnail

PhilippeWeidmann:
Commit 1: Fix: Upload wasn't triggered if the file already existed in DB
Commit 2: Apple doc says we should return nil if there is no thumbnail


adrien-coye:
Incremental update of a file works with `pages`. Merging ASAP.

PR #1277, opened 2024-09-11
fix(FileListViewController): Prevent realm crash in FileList

adrien-coye:
[edited] In some rare occurrence, the Drive object is invalidated and the `categories(for:` function breaks.
This is a workaround for that.


PhilippeWeidmann (changes requested):
Objects in `displayedFiles` are never invalidated because they are frozen. 
The crash occurs because we are trying to fetch live data in realm from the `FileListViewModel`. 
We need to check for invalidation before fetching categories. Maybe also check in other places in the viewmodel if we try to fetch live data.

adrien-coye:
Discussed with @PhilippeWeidmann The issue is actually with the `Drive` class. This PR now addresses the issue.

PhilippeWeidmann (approved):

PR #1279, opened 2024-09-11
fix(Deeplinking): Deeplink parsing is now migrated to Scene delegate

adrien-coye:
Also refactored the code to at least unify the navigation in one place and fix tests.


kDrive/AppRouter.swift
```diff
@@ -64,6 +65,45 @@ public struct AppRouter: AppNavigable {
         return userInfo
     }
 
+    // MARK: Routable
+
+    public func navigate(to route: NavigationRoutes) {
+        #if ISEXTENSION
```
adrien-coye: TODO remove, no longer needed.

kDriveCore/Utils/DeeplinkParser.swift
```diff
@@ -19,42 +19,48 @@
...
     @LazyInjectService var accountManager: AccountManageable
-    @LazyInjectService var navigationManager: NavigationManageable
+    @LazyInjectService var router: AppNavigable
 
     public init() {
         // META: keep SonarCloud happy
     }
 
-    public func parse(url: URL) -> Bool {
+    public func parse(url: URL) async -> Bool {
```
adrien-coye, line 41: Now async/await, only the rooting is on the main thread.

PhilippeWeidmann (approved):

PR #1281, opened 2024-09-12
fix: Legacy tabbar for iPadOS 18

PhilippeWeidmann:


adrien-coye (approved):
I did not ran the code or tested the fix, but did a quick review codewise. Looks good 👍

PR #1289, opened 2024-09-17
feat: Select all offline

PhilippeWeidmann:
Depends on #1287


adrien-coye (approved):
Code looks good, I'll try it out in the next beta.

PR #1292, opened 2024-09-17
fix: Prevent NSInternalInconsistencyException when reloading header

PhilippeWeidmann:
HomeView was crashing when showing UploadCell on some iOS versions. For some reason, `UICollectionViewLayoutInvalidationContext` is very hard and buggy to use. Switching from `invalidateDecorationElements` to `invalidateSupplementaryElements` seems to work without any side effect.
Fix for sentry short ids:
- KDRIVE-IOS-9Q7
- KDRIVE-IOS-8KX


adrien-coye (approved):
lets go !

PR #1293, opened 2024-09-17
fix: Realm accessed from the wrong thread

adrien-coye:
Sentry id KDRIVE-IOS-9PM


PhilippeWeidmann:
Not sure this will work every time. I think we have something fundamentally broken with notify/observe. Will discuss in DM.

PhilippeWeidmann (approved):

PR #1295, opened 2024-09-20
fix: Correctly observe parent directory for empty files

PhilippeWeidmann:
In some cases empty directory wasn't reported correctly. For example on initial opening of an empty directory.


adrien-coye (approved):
okay 👍

PR #1296, opened 2024-09-20
fix: Prevent selecting "Drive root" and select "My personal files"

PhilippeWeidmann:
We can't upload files to the drive root anymore so it doesn't make sense to select it by default. Personal files is instead selected.

### DIFF
--- a/kDrive/UI/Controller/Files/Save File/SaveFileViewController.swift
+++ b/kDrive/UI/Controller/Files/Save File/SaveFileViewController.swift
@@ -166,8 +166,8 @@ class SaveFileViewController: UIViewController {
 
         guard let selectedDriveFileManager else { return nil }
 
-        let myFilesDirectory = selectedDriveFileManager.database.fetchResults(ofType: File.self) { files in
-            files.filter(NSPredicate(format: "rawVisibility = %@", FileVisibility.isPrivateSpace.rawValue))
+        let myFilesDirectory = selectedDriveFileManager.database.fetchResults(ofType: File.self) { lazyFiles in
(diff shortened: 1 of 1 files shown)


kDrive/UI/Controller/Files/Save File/SaveFileViewController.swift
```diff
@@ -161,6 +159,30 @@ class SaveFileViewController: UIViewController {
...
     }
 
+    func getBestDirectory() -> File? {
+        if lastSelectedDirectory?.driveId == selectedDriveFileManager?.drive.id {
+            return lastSelectedDirectory
+        }
+
+        guard let selectedDriveFileManager else { return nil }
+
+        let myFilesDirectory = selectedDriveFileManager.database.fetchResults(ofType: File.self) { files in
```
adrien-coye, line 169: Maybe name it `lazyFiles in`

kDrive/UI/Controller/Files/Save File/SaveFileViewController.swift
```diff
@@ -161,6 +159,30 @@ class SaveFileViewController: UIViewController {
...
+            files.filter(NSPredicate(format: "rawVisibility = %@", FileVisibility.isPrivateSpace.rawValue))
```
adrien-coye, line 170: Tiny detail, we can write this directly.
```suggestion
            files.filter("rawVisibility = %@", FileVisibility.isPrivateSpace.rawValue)
```

adrien-coye (commented):
Looking good. Let me know what you think on my remarks and I'll approve it.

adrien-coye (approved):
ty

PR #1300, opened 2024-09-24
fix: Improve decoding of text files for preview

PhilippeWeidmann:


kDrive/UI/View/Files/Preview/CodePreviewCollectionViewCell.swift
```diff
@@ -67,7 +67,16 @@ class CodePreviewCollectionViewCell: PreviewCollectionViewCell {
     func configure(with file: File) {
         do {
             // Read file
-            let content = try String(contentsOf: file.localUrl)
+            let data = try Data(contentsOf: file.localUrl, options: .alwaysMapped)
+            var maybeString: NSString?
+
+            NSString.stringEncoding(for: data, convertedString: &maybeString, usedLossyConversion: nil)
```
adrien-coye, line 73: Yeah, even `String.Encoding` is bridged to older stuff, it looks like. 
Swift strings been designed to work with UTF in mind are not super useful working with pre Windows XP content. (UTF8 was supported but not the default on XP I believe)

adrien-coye (approved):
Did some testing with various files with broken encoding. Working well.
It looks like the work is done fully on the main thread. I'll suggest something in another PR.

PR #1301, opened 2024-09-24
fix(PreviewViewController): Read files outside of the main actor

adrien-coye:
When displaying a text file the UI was freezing.

IMG_3385

### DIFF
--- a/kDrive/UI/View/Files/Preview/CodePreviewCollectionViewCell.swift
+++ b/kDrive/UI/View/Files/Preview/CodePreviewCollectionViewCell.swift
@@ -40,14 +40,19 @@ struct CodePreviewWorker {
 
 class CodePreviewCollectionViewCell: PreviewCollectionViewCell {
     private let codePreviewWorker = CodePreviewWorker()
-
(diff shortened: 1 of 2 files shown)


kDrive/UI/View/Files/Preview/CodePreviewCollectionViewCell.swift
```diff
@@ -65,32 +101,37 @@ class CodePreviewCollectionViewCell: PreviewCollectionViewCell {
...
+            isCode = false
+        } else {
+            displayCode(for: content)
+            isCode = true
+        }
+    }
+
+    private func displayLoading() {
+        activityView.startAnimating()
+    }
```
PhilippeWeidmann, line 133: This function only calls `activityView.startAnimating()` and is called once so we don't really need it. Maybe we can simply call
`activityView.startAnimating()`  in the configure ?

PhilippeWeidmann (commented):
Juste one remark. Otherwise looks good 👍

PhilippeWeidmann (approved):

PR #1305, opened 2024-09-27
feat: Download content from External links

adrien-coye:
Functionalities of this PR.

- [x] Download file from Public Share to available drive.
- [x] Add to my drive
- [x] Download All


adrien-coye:
I did a clean PR instead.

PR #1306, opened 2024-09-27
feat: Public share feature branch

adrien-coye:
## Abstract

Feature PR of the external links

All the work so far:
- [x] Network calls
- [x] In memory realm for each share link
- [x] Propagate context to preview controller and subsequent `FileListViewController`
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1307
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1308
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1333
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1328
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1309
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1348
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1367
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1311
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1349
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1376
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1378
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1392
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1379
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1380
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1393
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1394
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1395
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1398

⚠ Must change deeplink handling backend side when the app is live.

PR #1311, opened 2024-09-27
feat: Multiselect behaviour

adrien-coye:
This PR makes `multi select` work with a public share.
Only one action, `Download all` is available in this mode.
`Select all` is also working with a dedicated route.

Sub PRs:
- [x] [Split archive operation](https://github.com/Infomaniak/ios-kDrive/pull/1368)
- [x] [Factorise a common DownloadOperation](https://github.com/Infomaniak/ios-kDrive/pull/1370)

Feature branch https://github.com/Infomaniak/ios-kDrive/pull/1306

IMG_3400


kDrive/AppRouter.swift
```diff
@@ -614,12 +614,27 @@ public struct AppRouter: AppNavigable {
                 return
             }
 
+            // TODO: i18n
```
adrien-coye, line 617: TODO i18n

kDrive/UI/Controller/Files/FilePresenter.swift
```diff
@@ -146,11 +146,23 @@ final class FilePresenter {
         if driveFileManager.drive.sharedWithMe {
             viewModel = SharedWithMeViewModel(driveFileManager: driveFileManager, currentDirectory: file)
         } else if let publicShareProxy = driveFileManager.publicShareProxy {
+            // TODO: i18n
```
adrien-coye, line 149: TODO: i18n

adrien-coye (commented):
Self assesment

kDriveCore/Data/DownloadQueue/DownloadArchiveOperation.swift
```diff
@@ -131,6 +146,12 @@ public class DownloadArchiveOperation: Operation {
         }
     }
 
+    func authenticatedDownload() {
+        DDLogInfo("[DownloadOperation] Downloading Archive of files \(archiveId) with session \(urlSession.identifier)")
+
+        // TODO: missing imp
```
adrien-coye, line 152: TODO

kDrive/UI/Controller/Files/MultipleSelectionFloatingPanelViewController.swift
```diff
@@ -139,7 +141,11 @@ final class MultipleSelectionFloatingPanelViewController: UICollectionViewContro
         }
     }
 
-    func downloadArchivedFiles(downloadCellPath: IndexPath, completion: @escaping (Result<URL, DriveError>) -> Void) {
+    // TODO: make it work
```
adrien-coye, line 144: TODO: Archive KO with public share

PhilippeWeidmann (commented):
Even if viewControllerDismissable is a protcol and is weak. I don't think it is a good practice to keep a reference to the ViewController from the ViewModel.

The VM shouldn't control the ViewController. However you indeed need to callback your VC if you want to dismiss it.
In the code base I see two examples: 
- Implement the logic inside the VC (see SearchViewController)
- Using a callback. It's more work but maybe more appropriate

adrien-coye:
@PhilippeWeidmann Yeah I agree, I felt it was not perfect while doing it. Yeah I can use a callback there.

PhilippeWeidmann (changes requested):
The DownloadQueue can handle different operations . 
We already have `DownloadOperation` and `DownloadArchiveOperation` maybe you could subclass `DownloadArchiveOperation` with `DownloadPublicShareArchiveOperation` it passing `shareDrive` and `publicShareProxy` when you don't need it. 

Also I think I missed it in an other review, but the same thing can be done for single files in `DownloadOperation` but the refactor should be done in a separate PR.

kDrive/UI/Controller/Files/File List/MultipleSelectionFileListViewModel.swift
```diff
@@ -23,35 +23,48 @@ import kDriveCore
...
+        case download
+    }
+
     static func == (lhs: MultipleSelectionAction, rhs: MultipleSelectionAction) -> Bool {
         return lhs.id == rhs.id
     }
 
     static let move = MultipleSelectionAction(
-        id: 0,
+        id: MultipleSelectionActionId.move,
```
PhilippeWeidmann, line 44: Same for the other ones
```suggestion
        id: .move,
```

kDrive/UI/Controller/Files/File List/FileListViewModel.swift
```diff
@@ -93,6 +93,7 @@ class FileListViewModel: SelectDelegate {
         var matomoViewPath = ["FileList"]
     }
 
+    var dismissClosure: (() -> Void)?
```
PhilippeWeidmann, line 96: There is already a naming pattern and a "region" to declare callbacks, see:
https://github.com/Infomaniak/ios-kDrive/blob/c35026c147d0c9811d2cde3990fbd55ab13a156d/kDrive/UI/Controller/Files/File%20List/FileListViewModel.swift#L136

kDriveCore/Data/DownloadQueue/DownloadQueue.swift
```diff
@@ -267,7 +296,7 @@ public final class DownloadQueue: ParallelismHeuristicDelegate {
     ///
     /// Thread safe
     /// Lookup O(1) as Dictionary backed
-    public func operation(for fileId: Int) -> DownloadOperation? {
+    public func operation(for fileId: Int) -> DownloadAuthenticatedOperation? {
```
adrien-coye, line 299: Should be the abstract `DownloadOperation` now

adrien-coye: Yes I can since last refactor. Using `DownloadFileOperationable` now.

PhilippeWeidmann (approved):

PR #1324, opened 2024-10-15
feat: Add now playing info for video

Matthieu-dgl:


kDrive/UI/View/Files/Preview/VideoCollectionViewCell.swift
```diff
@@ -123,6 +137,25 @@ class VideoCollectionViewCell: PreviewCollectionViewCell {
         floatingPanelController?.dismiss(animated: true)
         parentViewController?.present(navController, animated: true) {
             playerViewController.player?.play()
+            let interval = CMTime(seconds: 1.0, preferredTimescale: CMTimeScale(NSEC_PER_SEC))
+            player.addPeriodicTimeObserver(forInterval: interval, queue: .main) { [weak self] _ in
+                self?.setNowPlayingMetadata()
+            }
+        }
```
PhilippeWeidmann, line 144: In the `SingleTrackPlayer` we don't call `setNowPlayingMetadata` to update time. Is it really needed ?

PhilippeWeidmann (changes requested):
You only observe `.AVPlayerItemDidPlayToEndTime`.
This only set the correct metadata at the end of the track.

Maybe try to also remove  `MPMediaItemPropertyPlaybackDuration` and `MPNowPlayingInfoPropertyElapsedPlaybackTime` because they should be set by the VideoPlayer

Matthieu-dgl:
There's always the problem that the video title appears for 1 second and then disappears in the notfication. I'm open to an idea!

kDrive/UI/View/Files/Preview/VideoCollectionViewCell.swift
```diff
@@ -125,4 +144,29 @@ class VideoCollectionViewCell: PreviewCollectionViewCell {
             playerViewController.player?.play()
         }
     }
+
+    private func setNowPlayingMetadata() {
```
adrien-coye, line 148: Architecture wise, it is complex enough to justify a `SingleVideoPlayer` struct to encapsulate the media player observation.

You can copy what was done with `SingleTrackPlayer` for the link between the view and the player object.

adrien-coye (changes requested):
I'll check the `1 second and then disappears`  asap.

kDriveCore/VideoPlayer/VideoPlayer.swift
```diff
@@ -0,0 +1,104 @@
...
+import Combine
+import FloatingPanel
+import InfomaniakCore
+import kDriveCore
+import kDriveResources
+import Kingfisher
+import MediaPlayer
+import UIKit
+
+class VideoPlayer {
```
adrien-coye, line 29: Since `VideoPlayer` is now in `Core` you need to make it a public type.
You could build a protocol with all the public properties you want to use from the outside.

kDrive/UI/View/Files/Preview/VideoCollectionViewCell.swift
```diff
@@ -66,37 +62,27 @@ class VideoCollectionViewCell: PreviewCollectionViewCell {
...
     }
 
     override func didEndDisplaying() {
-        MatomoUtils.trackMediaPlayer(leaveAt: player?.progressPercentage)
+        MatomoUtils.trackMediaPlayer(leaveAt: videoPlayer?.progressPercentage)
     }
 
     @IBAction func playVideoPressed(_ sender: Any) {
-        guard let player else { return }
+        guard let player = videoPlayer?.avPlayer else {
```
adrien-coye, line 83: I'm not found of exposing directly the AVPlayer in the cell. 
It goes against encapsulation, I would rather use a mechanism for decoupling (like Combine in the SingleTrackPlayer)

kDrive/UI/View/Files/Preview/VideoCollectionViewCell.swift
```diff
@@ -110,17 +96,26 @@ class VideoCollectionViewCell: PreviewCollectionViewCell {
...
+        }
+    }
+
+    private func presentFloatingPanel() {
+        if let floatingPanelController = parentViewController?.presentedViewController as? FloatingPanelController {
+            floatingPanelController.dismiss(animated: true)
+        }
+    }
+
+    private func presentVideoPlayer(navController: VideoPlayerNavigationController,
```
adrien-coye, line 117: It looks like it is an unused method. Should we remove it ?

adrien-coye (changes requested):
I have a couple of small remarks regarding the `VideoPlayer`.

I don't think your branch can build right now. We usually try to have branches that can build even if the app is broken.

adrien-coye:
As a "nice to have", we could also set a thumbnail. We already have one, it should be straight forward.

kDrive/UI/View/Files/Preview/VideoCollectionViewCell.swift
adrien-coye: Discussed, you can use this in `VideoPlayer` to remove the property `avPlayer`

```swift
    public lazy var playerViewController: AVPlayerViewController = {
        let playerViewController = AVPlayerViewController()
        playerViewController.player = self.player
        return playerViewController
    }()
```

kDriveCore/VideoPlayer/VideoPlayer.swift
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    private func addPeriodicTimeObserver() {
+        guard let player = player else { return }
+        let interval = CMTime(seconds: 1, preferredTimescale: CMTimeScale(NSEC_PER_SEC))
+        player.addPeriodicTimeObserver(forInterval: interval, queue: .main) { [weak self] _ in
+            self?.updateNowPlayingInfo()
+        }
+    }
```
PhilippeWeidmann, line 105: There is probably a better way to connect AVPlayer's time to the MPNowPlayingInfoCenter

adrien-coye (commented):
Looking at the code, it now looks like the `VideoPlayer` is decoupled enough. I no longer see `AVPlayer` in the view. Thanks for working on this @Matthieu-dgl .

I'll test the PR again later, including the audio player, when [the change](https://github.com/Infomaniak/ios-kDrive/pull/1324/files#r1849827164) asked by @PhilippeWeidmann is done.

kDriveCore/VideoPlayer/VideoPlayer.swift
adrien-coye: I'll check the PR again after this is done.

adrien-coye: We checked with @Matthieu-dgl The alternative solution seems to be using KVO.

kDrive/UI/View/Files/Preview/VideoCollectionViewCell.swift
```diff
@@ -41,13 +42,10 @@ class VideoCollectionViewCell: PreviewCollectionViewCell {
     weak var parentViewController: UIViewController?
     weak var floatingPanelController: FloatingPanelController?
 
+    private var playableFileName: String?
```
adrien-coye, line 45: Looks like it is unused.

kDriveCore/AudioPlayer/MediaMetadata.swift
```diff
@@ -0,0 +1,64 @@
...
+
+    public init(title: String, artist: String, artwork: UIImage?) {
+        self.title = title
+        self.artist = artist
+        self.artwork = artwork
+    }
+}
+
+public extension MediaMetadata {
+    static func extractTrackMetadata(from url: URL) async -> MediaMetadata {
```
adrien-coye, line 36: pass a nullable `playableFileName`

kDriveCore/AudioPlayer/MediaMetadata.swift
```diff
@@ -0,0 +1,64 @@
...
+        let asset = AVAsset(url: url)
+
+        var title = KDriveResourcesStrings.Localizable.unknownTitle
```
adrien-coye, line 39: Something like this for the audio tracks
`var title = playableFileName ?? KDriveResourcesStrings.Localizable.unknownTitle`

adrien-coye (commented):
We checked the code together, I need to play with files with metadata.

kDriveCore/VideoPlayer/VideoPlayer.swift
adrien-coye: I took some time to spend on this issue.
Yeah unlike for an audio track where we need to track progress for our _own_ view, we can do without KVO and without a `PeriodicTimeObserver`. Also for the audio player, a nice to have could be to use a `CADisplayLink` instead.

kDriveCore/AudioPlayer/MediaMetadata.swift
```diff
@@ -0,0 +1,62 @@
...
+    public init(title: String, artist: String, artwork: UIImage?) {
+        self.title = title
+        self.artist = artist
+        self.artwork = artwork
+    }
+}
+
+public extension MediaMetadata {
+    static func extractTrackMetadata(from url: URL, playableFileName: String?) async -> MediaMetadata {
+        let asset = AVAsset(url: url)
```
adrien-coye, line 37: The regression on the audio part is that an `AVAsset` will not work with a remote URL.
When trying to stream content, the single track audio player relies on a `AVURLAsset` instead.
This needs to change back, we can use this opportunity to abstract the AV types if you prefer.

kDriveCore/AudioPlayer/MediaPlayerOrchestrator.swift
```diff
@@ -18,23 +18,33 @@
...
 import Foundation
 
+public protocol Pausable {
+    func pause()
+}
+
 /// A simple service to orchestrate between all the possibly existing media player objects in the app
-public final class MediaPlayerOrchestrator {
+public final class MediaPlayerOrchestrator: Pausable {
     private weak var lastUsedPlayer: SingleTrackPlayer?
```
adrien-coye, line 27: This could be a `Pausable` too

adrien-coye (commented):
Thanks for applying changes, I'll show you tomorrow a _patch_ to finish the last details.

kDriveCore/AudioPlayer/MediaPlayerOrchestrator.swift
adrien-coye: discussed

adrien-coye (approved):
Thanks for the changes. It looks good to me, I'll ask @PhilippeWeidmann  for a final review.

PhilippeWeidmann (approved):

PR #1325, opened 2024-10-16
fix: Multiple selection action not same between iOS and Android, plus fix color

Matthieu-dgl:
… fix color

### DIFF
--- a/kDriveCore/Data/Cache/DriveFileManager/DriveFileManager.swift
+++ b/kDriveCore/Data/Cache/DriveFileManager/DriveFileManager.swift
@@ -43,7 +43,7 @@ public final class DriveFileManager {
     }
 
     public static var sharedWithMeRootFile: File {
-        return File(id: -3, name: "Shared with me")
+        return File(id: -10, name: "Shared with me", visibility: .isSharedSpace)
     }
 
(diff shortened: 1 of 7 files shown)


kDriveCore/Data/Cache/DriveFileManager/DriveFileManager.swift
```diff
@@ -72,6 +68,11 @@ public final class DriveFileManager {
         return offlineRoot
     }
 
+    public static var sharedWithMeRootFile: File {
+        // Default value is -3
```
PhilippeWeidmann, line 72:
```suggestion
        // We can't migrate fake roots. Previous sharedWithMeRootFile.id was -3
```

kDriveCore/Data/Models/Rights.swift
```diff
@@ -73,6 +75,7 @@ public class Rights: EmbeddedObject, Codable {
         case canUpload
         case canMoveInto
         case canBecomeDropbox
+        case canColor
```
PhilippeWeidmann, line 78: canColor key doesn't exist in the json
```suggestion
        case canColor = "colorable"
```

PhilippeWeidmann (approved):

PR #1330, opened 2024-10-25
chore: Fetch i18n to date

adrien-coye:
Bump i18n to the latest version of texts that include "public shares"


adrien-coye:
@valentinperignon approved in DM

PR #1332, opened 2024-10-28
chore: Fetch i18n to date

adrien-coye:
Reverts Infomaniak/ios-kDrive#1331


adrien-coye:
Made a clean PR instead of this one.

PR #1336, opened 2024-11-01
fix: Handle AVPlayer.currentItem.status error

PhilippeWeidmann:
Depends on #1345


adrien-coye (commented):
Reviewed the code, it does look fine to me.
I'll merge it after testing with a FLAC file tomorow.

adrien-coye (approved):
Discussed, we are happy with this.

PR #1337, opened 2024-11-06
fix: Cache cleanup fix

adrien-coye:
- Each chunk is now cleaned right away after a successful upload. In some heavy system load conditions, the clean-up was not performed.
- The clean-up of the cache folder is triggered earlier on
- The cleanup is not only done when the app closes, but now also before we try to rebuild the upload queue

### DIFF
--- a/kDriveCore/Data/UploadQueue/Operation/UploadOperation+UploadChunk.swift
+++ b/kDriveCore/Data/UploadQueue/Operation/UploadOperation+UploadChunk.swift
@@ -98,7 +98,7 @@ extension UploadOperation {
 
         // Write buffer
         let chunkName = chunkName(number: number, fileId: uploadFileId, hash: hash)
-        let chunkPath = tempChunkFolder.appendingPathExtension(chunkName)
+        let chunkPath = tempChunkFolder.appendingPathComponent(chunkName)
         try buffer.write(to: chunkPath, options: [.atomic])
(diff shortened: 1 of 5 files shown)


kDriveCore/Data/UploadQueue/Operation/UploadOperation+UploadChunk.swift
```diff
@@ -87,18 +87,16 @@ extension UploadOperation {
...
+                    sessionToken: String,
                     hash: String) throws -> URL {
-        // Create subfolders if needed
-        let tempChunkFolder = buildFolderPath(fileId: uploadFileId, sessionToken: sessionToken)
-        Log.uploadOperation("using chunk folder:'\(tempChunkFolder)' ufid:\(uploadFileId)")
-        if !fileManager.fileExists(atPath: tempChunkFolder.path, isDirectory: nil) {
-            try fileManager.createDirectory(at: tempChunkFolder, withIntermediateDirectories: true, attributes: nil)
-        }
+        // Store chunks at the root of NSTemporaryDirectory
+        let tempRoot = URL(fileURLWithPath: NSTemporaryDirectory())
```
valentinperignon, line 96: Why don't you use `FileManager.default.temporaryDirectory` ?

adrien-coye: Yeah good point. Make sense to get an URL directly there.

PR #1339, opened 2024-11-08
fix: Pictures no longer on device but only on iCloud should be uploaded as well.

adrien-coye:
- Correct error handling during the hashing of a photo source.
- Sending an event to sentry if an error is raised when we try to fetch an asset that might only be available remotely on iCloud.

I do not skip assets that we are unable to access, as this could result in a loss of data for the consumer.


kDriveCore/Utils/PHAsset/PHAssetIdentifier.swift
```diff
@@ -147,32 +147,43 @@ struct PHAssetIdentifier: PHAssetIdentifiable {
...
                 Log.photoLibraryUploader("hashing resource \(progress * 100)% …")
             }
 
             group.enter()
+            var resourceManagerError: Error?
             PHAssetResourceManager.default().requestData(for: bestResource,
                                                          options: options) { data in
                 hasher.update(data)
-            } completionHandler: { _ in
+            } completionHandler: { error in
```
adrien-coye, line 161: I now capture the error if we fail to get an asset to hash it during the library scanning.

valentinperignon (approved):

PR #1342, opened 2024-11-26
fix: Free accounts can see change color action

adrien-coye:
Free accounts will be presented with an upsale sheet.

### DIFF
--- a/kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController+Actions.swift
+++ b/kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController+Actions.swift
@@ -72,8 +72,10 @@ extension FileActionsFloatingPanelViewController {
                 return file.capabilities.canBecomeDropbox
             case .manageDropbox:
                 return file.isDropbox
+            case .upsaleColor:
+                return file.isDirectory && driveFileManager.drive.isFreePack
             case .folderColor:
-                return file.isDirectory // True for directory to display an upSaleView is necessary
+                return file.capabilities.canColor
             case .seeFolder:
                 return !normalFolderHierarchy && (file.parent != nil || file.parentId != 0)
             case .offline:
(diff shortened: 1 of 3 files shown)


kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController+Actions.swift
```diff
@@ -73,7 +73,7 @@ extension FileActionsFloatingPanelViewController {
             case .manageDropbox:
                 return file.isDropbox
             case .folderColor:
-                return file.capabilities.canColor
```
valentinperignon, line 76: suggestion: I think we should have a more complex condition here.
If the user has a free pack, then we can show the option if `file.isDirectory`, else we should listen to `file.capabilities.canColor`.

adrien-coye: I hear you, but I think you over complexity things.
This logic enables the action in the view if it is a directory, it does not check the capabilities.
Then performing the action, a capability check is performed. And we either display an up-sale sheet if the capability is not there, or we perform the action. Returning false here will hide the action. The way it was built before me is this one I believe.

adrien-coye: Maybe I could rename the Action to explicit that it will check capabilities on its own.

adrien-coye: You can also check in `MultipleSelectionFloatingPanelViewController` where a similar logic is performed, there is zero check on capability.

kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController+Actions.swift
```diff
@@ -72,8 +72,10 @@ extension FileActionsFloatingPanelViewController {
                 return file.capabilities.canBecomeDropbox
             case .manageDropbox:
                 return file.isDropbox
+            case .upsaleColor:
+                return file.isDirectory && driveFileManager.drive.isFreePack
             case .folderColor:
-                return file.capabilities.canColor
+                return file.capabilities.canColor && !driveFileManager.drive.isFreePack
```
adrien-coye, line 78: Something I dislike here is that without a check for the free pack, I got a second identical action displayed if we rely only on the `canColor` of the backend.

PR #1343, opened 2024-11-26
fix: Prevent a realm crash when saving a file

adrien-coye:
A deeper refactor could be performed to explicit the use of frozen files in `File actions`.
Right now this will prevent a realm crash when using the `save` action.


PhilippeWeidmann (approved):
I agree, this will do for now but the current code has lot of room for improvements.

PR #1344, opened 2024-11-27
fix: Correctly select elements in gallery

PhilippeWeidmann:
`FileListViewController` uses `displayedFiles` which is an array because there are no sections. 
`PhotoListViewController` extends `FileListViewController` but has a different datasource implementation with one section per month. This creates a mixup of indexes on selection.

This PR adds a `getDisplayedFile(at indexPath:IndexPath)` that can be overridden by subclasses for custom behaviour


adrien-coye (approved):
I played with the code a bit without issues. Thanks for the informations about the `deleteOrphanFiles` behaviour.

PR #1346, opened 2024-12-05
fix(AudioCollectionViewCell): Correctly import IKLabel

PhilippeWeidmann:
IKLabel wasn't imported correctly in XIB and no text style was set.


adrien-coye (approved):
I tested it on Ip8. Works fine with me.

PR #1351, opened 2024-12-13
refactor(FileActionsFloatingPanelViewController): Using an enum

adrien-coye:
Using an enum for identifiers to prevent bugs.


kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController.swift
```diff
@@ -24,6 +24,31 @@ import kDriveResources
 import LinkPresentation
 import UIKit
 
+public enum FloatingPanelActionIdentifier: Int {
```
PhilippeWeidmann, line 27: We could use an enum with no raw value and have `FloatingPanelAction.id: FloatingPanelActionIdentifier` to prevent passing simply an Int. 

What do you think ?

```suggestion
public enum FloatingPanelActionIdentifier {
```

kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController.swift
PhilippeWeidmann, line 1: While we are refactoring. We could take the opportunity to move `FloatingPanelAction` to it's own file (with FloatingPanelActionIdentifier inside as it doesn't need to be public)

kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController.swift
adrien-coye: I went for a minimum change / max impact kind of change, thus no change on `FloatingPanelAction`.

Yeah this does removes a class of issues dealing with `int` directly when switching on this.

kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController.swift
adrien-coye: sure 👍

PhilippeWeidmann (commented):
All good

PhilippeWeidmann (approved):

PR #1353, opened 2024-12-13
refactor(SaveFileViewController): Split the class

adrien-coye:
Split the class for better readability and further development.
Strictly compatible with the current code, does not change anything.
Required by `PublicShare` feature.


PhilippeWeidmann (changes requested):
We used // MARK for easier navigation but it isn't needed anymore since the code is now split in multiple files

PhilippeWeidmann (approved):

PR #1354, opened 2024-12-13
fix(PhotoLibraryUploader): Early photo upload start.

adrien-coye:
Start the upload during a photo sync after the scan of the first batch of pictures.
Strictly limited to foreground app operations.


adrien-coye:
If the upload starts before the initial scan is performed, it can become flaky.
The trivial patch is not sufficient as is, moved to Draft again.

adrien-coye:
Non trivial to fix. Closing this PR for now to remove noise.

PR #1370, opened 2025-01-02
refactor: Common download operation

adrien-coye:
We can push the refactor further and share the boilerplate of all `DownloadOperations`.

I did not rewrite old code that might need further rewrite. This is just code factorisation.

Intended to be reviewed per commit.

Root PR: https://github.com/Infomaniak/ios-kDrive/pull/1311


PhilippeWeidmann (commented):
LGTM.

One question, I'm a bit on the fence regarding the `@unchecked Sendable` for:
- DownloadOperation
- DownloadArchiveOperation

The other operations are fine because they don't have additional mutable state. However in the case of `DownloadArchiveOperation` we rely on the task that urlSession _should_ be thread safe and calls in a queue `downloadCompletion` and assigns `archiveUrl` but nothing protects archiveUrl from being assigned elsewhere. 

I won't request changes but this is something we have to keep in mind when using @unchecked

PhilippeWeidmann (approved):

adrien-coye:
Discussed regarding the `@unchecked Sendable` used with legacy Obj-c code.
Do not use it with legacy code without an audit, else do not use it and keep a warning, which make sense.

PR #1349, opened 2024-12-06
feat: Add to my drive action

adrien-coye:
Feature PR https://github.com/Infomaniak/ios-kDrive/pull/1306

Making sure to link the last details together, including a proper navigation.
- [x] Login Logout behaviour
- [x] Add to my kDrive main button
- [x] Make sure some calls are not blocked by an unnecessary auth token
- [x] Review navigation, must work logged in or out (AppRouter)
- [x] Streaming is disabled
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1358 for post authentication actions
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1350
- [x] https://github.com/Infomaniak/ios-kDrive/pull/1375

PR #1375, opened 2025-01-09
feat: External links download all progres UI

adrien-coye:
Display an activity indicator while the download proceeds.
Can cancel a download all action with a tap.

root pr https://github.com/Infomaniak/ios-kDrive/pull/1349

### DIFF
--- a/kDrive/UI/Controller/Menu/Share/PublicShareViewModel.swift
+++ b/kDrive/UI/Controller/Menu/Share/PublicShareViewModel.swift
@@ -103,12 +103,16 @@ final class PublicShareViewModel: InMemoryFileListViewModel {
     private func downloadAll(sender: Any?, publicShareProxy: PublicShareProxy) {
         let button = sender as? UIButton
         button?.isEnabled = false
+        configuration.rightBarButtons = [.loading]
+        loadButtonsConfiguration()
 
         downloadObserver = DownloadQueue.instance
             .observeFileDownloaded(self, fileId: currentDirectory.id) { [weak self] _, error in
                 Task { @MainActor in
                     defer {
                         button?.isEnabled = true
+                        self?.configuration.rightBarButtons = [.downloadAll]
+                        self?.loadButtonsConfiguration()
                     }
 
                     guard let self = self else {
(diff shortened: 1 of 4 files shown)


kDrive/UI/Controller/Menu/Share/PublicShareViewModel.swift
```diff
@@ -100,24 +101,34 @@ final class PublicShareViewModel: InMemoryFileListViewModel {
...
+        DownloadQueue.instance.cancelFileOperation(for: currentDirectory.id)
+        clearDownloadObserver()
+        configuration.rightBarButtons = [.downloadAll]
+        loadButtonsConfiguration()
+    }
+
     private func downloadAll(sender: Any?, publicShareProxy: PublicShareProxy) {
         let button = sender as? UIButton
         button?.isEnabled = false
+        configuration.rightBarButtons = [.downloadingAll]
```
adrien-coye, line 114: IMO it is cleaner to change the configuration and refresh than write directly to the ViewModel state.

kDrive/UI/View/Files/FileListBarButton.swift
```diff
@@ -53,11 +53,26 @@ final class FileListBarButton: UIBarButtonItem {
...
+        case .downloadingAll:
+            self.init(title: nil, style: .plain, target: target, action: action)
+
+            let activityView = UIActivityIndicatorView(style: .medium)
+            activityView.startAnimating()
+
+            let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(cancelDownloadPressed))
+            activityView.addGestureRecognizer(tapGestureRecognizer)
+
+            customView = activityView
```
adrien-coye, line 65: Having an action matching our pattern of calling on a UITabBarItem with a custom view was not trivial.

adrien-coye (commented):
left some notes for reviewer

PhilippeWeidmann (approved):

kDrive/UI/View/Files/FileListBarButton.swift
PhilippeWeidmann: In the end, that makes sense as we are replacing a button with something that's not a buttob

PR #1383, opened 2025-01-14
fix(PhotoSync): Skip iCloud asset on iCloud critical error

adrien-coye:
- PhotoSync will skip an iCloud asset if Apple's API is in error. The scan will continue.
- Small changes on `Sentry` related code for better tracking of the issue.


adrien-coye:
I did some testing with an iPhone with iCloud sync. I also did some testing, injecting errors during an initial photo scan.
Both cases worked without issues.

PhilippeWeidmann (approved):

PR #1388, opened 2025-01-20
refactor(UploadQueueViewController): Using DifferenceKit to prevent a crash

adrien-coye:


PhilippeWeidmann (changes requested):
It's working great ! 

You are missing corners for diffing you can take a look at how it's done here:
https://github.com/Infomaniak/ios-kDrive/blob/3ee5a3b882bbf22adc8731cf7559bb764fb08980/kDriveCore/Data/Models/File.swift#L448

But it's not very good to have it here. I think we could do a container type that looks like this now that we have the same use in 3 places.

Something like 
```swift
struct CornerCellContainer<Content: Differentiable> : Differentiable
let isFirstInList: Bool
let isLastInList: Bool
let content: Content
```

Let's discuss it irl.

PhilippeWeidmann (approved):

PR #1394, opened 2025-01-22
fix: Public share link behaviour on login and logout

adrien-coye:


kDrive/UI/Controller/LoginDelegateHandler.swift
```diff
@@ -64,6 +65,10 @@ public final class LoginDelegateHandler: InfomaniakLoginDelegate {
         UserDefaults.shared.legacyIsFirstLaunch = false
         UserDefaults.shared.numberOfConnections = 1
         _ = router.showMainViewController(driveFileManager: driveFileManager, selectedIndex: nil)
+
+        Task {
+            deeplinkService.processDeeplinksPostAuthentication()
+        }
```
PhilippeWeidmann, line 71: This looks weird, Task with no await and no MainActor ? Also `processDeeplinksPostAuthentication` already has a Task inside implementation.

PhilippeWeidmann (approved):

PR #1410, opened 2025-01-28
fix(DirectoryEnumerator): Remove deleted items from updated items

PhilippeWeidmann:
fix for KDRIVE-IOS-BWQ

If in the same action result the user has updated and deleted a file (eg. rename then delete). The fileprovider would crash because the updated file was invalidated. 

This PR removes the file from the updatedItems if its also in the deletedItems


adrien-coye (approved):
Code looks good.
I did a rename + delete, no crash to report.

PR #1412, opened 2025-01-30
fix: Invalidated live object crash DiffKit while diffing

adrien-coye:
### Abstract 

When cancelling an upload, the app crashes.

`DiffKit` accesses an invalidated object and crashes.
The `UploadQueueViewController` now tracks a list of frozen files.
The cell was updated to use frozen files as input.
The use of frozen or live file was explicited

### DIFF
--- a/kDrive/UI/View/Files/Upload/UploadTableViewCell.swift
+++ b/kDrive/UI/View/Files/Upload/UploadTableViewCell.swift
@@ -104,8 +104,9 @@ final class UploadTableViewCell: InsetTableViewCell {
         }
     }
 
-    func configureWith(uploadFile: UploadFile, progress: CGFloat?) {
-        guard !uploadFile.isInvalidated else {
+    func configureWith(frozenUploadFile: UploadFile, progress: CGFloat?) {
+        assert(frozenUploadFile.isFrozen, "Expected a frozen upload file")
+        guard let uploadFile = frozenUploadFile.thaw() else {
             return
         }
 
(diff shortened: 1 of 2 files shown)


adrien-coye:
Did some manual testing, it does fix the issue. I did not encouter a performance issue.

kDrive/UI/View/Files/Upload/UploadTableViewCell.swift
```diff
@@ -104,8 +104,9 @@ final class UploadTableViewCell: InsetTableViewCell {
         }
     }
 
-    func configureWith(uploadFile: UploadFile, progress: CGFloat?) {
-        guard !uploadFile.isInvalidated else {
+    func configureWith(frozenUploadFile: UploadFile, progress: CGFloat?) {
+        assert(frozenUploadFile.isFrozen, "Expected a frozen upload file")
+        guard let uploadFile = frozenUploadFile.thaw() else {
```
PhilippeWeidmann, line 109: The user will end up with an incorrect cell because we don't reconfigure it after reuse. It prevents the crash but the state of the cell will be incorrect.

This was also wrong before, we shouldn't have merged the `uploadFile.isInvalidated` in the first place. 
For now I accept the changes but we shouldn't do simple invalidated checks in the future.

PhilippeWeidmann (approved):

kDrive/UI/View/Files/Upload/UploadTableViewCell.swift
adrien-coye: The scenario where the crash typically occurs is when an upload is done and the entry is removed from the list.
The scenario you describe is technically correct but rare in actuality.

PR #1414, opened 2025-01-30
chore: Bump core

PhilippeWeidmann:


kDriveCore/Data/Api/PublicShareApiFetcher.swift
```diff
@@ -30,15 +30,16 @@ public enum PublicShareLimitation: String {
 }
 
 public class PublicShareApiFetcher: ApiFetcher {
-    override public init() {
+    public init() {
```
adrien-coye, line 33: Do we still need the init? (since it only calls super now)

adrien-coye (commented):
Did a quick read of the _draft_ code to understand the implications of the underlying core lib.

adrien-coye (approved):
LGTM, also did a quick test.

PR #1415, opened 2025-01-30
chore: Set minimum target to iOS15

adrien-coye:


kDrive/UI/Controller/Create File/PlusButtonFloatingPanelViewController.swift
```diff
@@ -241,7 +248,9 @@ class PlusButtonFloatingPanelViewController: UITableViewController, FloatingPane
     // MARK: Actions
 
     private func importAction(_ mainTabViewController: MainTabViewController) {
-        let documentPicker = DriveImportDocumentPickerViewController(documentTypes: [UTI.data.identifier], in: .import)
+        guard let utType = UTI.data.asUTType else { return }
```
adrien-coye, line 251: This is less than perfect, a lengthy `UTI` to `UTType` migration should be done both in `kDrive` and `ios-core`
This minimalist compatibility layer allows us to bump to iOS15 min without deprecation warnings and without a long refactor.

adrien-coye:
Each and every API update was tested for iOS 14 migrations. I'm confident with this work.

Tuist/ProjectDescriptionHelpers/Constants.swift
```diff
@@ -28,7 +28,7 @@ public enum Constants {
         .currentProjectVersion("1")
         .marketingVersion("5.1.0")
 
-    public static let deploymentTarget = DeploymentTargets.iOS("13.4")
+    public static let deploymentTarget = DeploymentTargets.iOS("15.0")
```
adrien-coye, line 31: @valentinperignon Is there a value SwiftUI wise to target 15.1 instead ? Or will this do nicely as a base target ?

adrien-coye: Discussed, mail is targeting 15.0. We stick to this then.

kDrive/AppDelegate.swift
```diff
@@ -163,7 +163,7 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
                                 withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
         _ = notification.request.content.userInfo
 
-        completionHandler([.alert, .sound])
+        completionHandler([.list, .banner, .sound])
```
adrien-coye, line 166: Migrated to something strictly equivalent

kDrive/UI/Controller/Create File/PlusButtonFloatingPanelViewController.swift
```diff
@@ -241,7 +248,9 @@ class PlusButtonFloatingPanelViewController: UITableViewController, FloatingPane
...
+        let documentPicker = DriveImportDocumentPickerViewController(forOpeningContentTypes: [utType])
```
PhilippeWeidmann, line 252: We could simply do this instead of adding a translation layer ?
```suggestion
        let documentPicker = DriveImportDocumentPickerViewController(forOpeningContentTypes: [UTType.data])
```

PhilippeWeidmann (approved):

PR #1420, opened 2025-02-04
fix: More defensive enum decoding

PhilippeWeidmann:


adrien-coye (commented):
Code looks good, I want to play with it a bit before giving the final OK.

adrien-coye (approved):
I did test the app, including public share. All good 👍

PR #1422, opened 2025-02-07
chore: Upgrade tuist + Correctly declare dependencies

PhilippeWeidmann:


adrien-coye (commented):
App runs fine, as well as the share extension.

adrien-coye (commented):
Onboarding works fine 👍

PR #1429, opened 2025-02-12
feat(FloatingPanelActionCollectionViewCell): Chip capability

adrien-coye:
IMG_3449
IMG_3448

### DIFF
--- a/kDrive/UI/View/Files/FloatingPanel/FloatingPanelActionCollectionViewCell.swift
+++ b/kDrive/UI/View/Files/FloatingPanel/FloatingPanelActionCollectionViewCell.swift
@@ -76,8 +76,8 @@ class FloatingPanelActionCollectionViewCell: UICollectionViewCell {
         }
 
         switch action {
-        case .upsaleColor:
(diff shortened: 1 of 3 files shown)


kDrive/UI/View/Files/FloatingPanel/FloatingPanelActionCollectionViewCell.swift
```diff
@@ -70,6 +74,14 @@ class FloatingPanelActionCollectionViewCell: UICollectionViewCell {
                 observeProgress(showProgress, file: file)
             }
         }
+
+        switch action {
+        case .upsaleColor, .convertToDropbox:
```
PhilippeWeidmann, line 79: We should also check quota for convertToDropbox because we can have 1 dropbox per free account

adrien-coye: This is only a UI pr, I do not add quota checks in this PR (nor the chip one)

PhilippeWeidmann (approved):

PR #1434, opened 2025-02-14
fix(AppRouter): Prevent app restoration if Drive are not matching

adrien-coye:


adrien-coye:
superseeded https://github.com/Infomaniak/ios-kDrive/pull/1437

adrien-coye:
superseeded https://github.com/Infomaniak/ios-kDrive/pull/1437

PR #1435, opened 2025-02-17
Fix deeplink navigation

adrien-coye:


adrien-coye:
superseeded https://github.com/Infomaniak/ios-kDrive/pull/1436

PR #1436, opened 2025-02-17
fix: Scene deeplink navigation

adrien-coye:


kDrive/SceneDelegate.swift
```diff
@@ -222,6 +229,17 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, AccountManagerDel
         }
     }
 
+    @discardableResult
+    private func continueToWebActivityIfPossible(_ scene: UIScene, userActivity: NSUserActivity) async -> Bool {
```
PhilippeWeidmann, line 233: In the `continueToWebActivityIfPossible` the `to` implies that we go to the web. Instead we are coming from the web and we try to `continueWebActivity`

```suggestion
    private func continueWebActivityIfPossible(_ scene: UIScene, userActivity: NSUserActivity) async -> Bool {
```

kDrive/SceneDelegate.swift
```diff
@@ -222,6 +229,17 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, AccountManagerDel
...
+        guard userActivity.activityType == NSUserActivityTypeBrowsingWeb,
+              let incomingURL = userActivity.webpageURL else {
+            Log.sceneDelegate("the scene continue userActivity - is not NSUserActivityTypeBrowsingWeb", level: .error)
+            return false
+        }
+
+        return await DeeplinkParser().parse(url: incomingURL)
```
PhilippeWeidmann, line 240: We can use the injection now. There is also an other occurence in class SceneDelegate.swift:227
```suggestion
        return await deeplinkParser.parse(url: incomingURL)
```

adrien-coye: I changed all remaining occurrences of this

PhilippeWeidmann (approved):

adrien-coye:
I did some deep link validation tests:
- onlyOffice file with correct user logged in { app opened, closed }
- onlyOffice file with incorrect user logged in { app opened, closed }
- public share user logged in { app opened, closed }
- public share logged out { app opened, closed }

Behaviour is OK, merging now …

PR #1437, opened 2025-02-17
fix: Drive restoration to the correct one

adrien-coye:
TODO: Use currentDriveId

- Restoration in not performed if Drive mismatch.
- If restoration data available, we select the last used drive.


adrien-coye:
Superseded by https://github.com/Infomaniak/ios-kDrive/pull/1442

PR #1439, opened 2025-02-17
fix: DriveFileManager.drive returns an object up to date

adrien-coye:


kDriveCore/Data/Cache/DriveFileManager/DriveFileManager.swift
```diff
@@ -90,8 +91,23 @@ public final class DriveFileManager {
...
     public let realmConfiguration: Realm.Configuration
 
-    // TODO: Fetch a drive with a computed property instead of tracking a Realm object
-    public private(set) var drive: Drive
+    private var _drive: Drive
+    public var drive: Drive {
+        guard let freshFrozenDrive = driveInfoDatabase.fetchObject(ofType: Drive.self, filtering: { lazyCollection in
+            return lazyCollection
+                .filter("id = %lld", driveId)
+                .first?
```
PhilippeWeidmann, line 99: We can use _drive.objectId to avoid filtering on collection

PhilippeWeidmann (approved):

PR #1440, opened 2025-02-17
feat: PhotoSync with explicit wifi or 4G

BaptGrv:
Depends on https://github.com/Infomaniak/ios-kDrive/pull/1444
Depends on https://github.com/Infomaniak/ios-kDrive/pull/1441
Depends on https://github.com/Infomaniak/ios-kDrive/pull/1452
Depends on https://github.com/Infomaniak/ios-kDrive/pull/1453
Depends on https://github.com/Infomaniak/ios-kDrive/pull/1460
Depends on https://github.com/Infomaniak/ios-kDrive/pull/1463
~~Depends on https://github.com/Infomaniak/ios-kDrive/pull/1466~~
Depends on https://github.com/Infomaniak/ios-kDrive/pull/1470
Depends on https://github.com/Infomaniak/ios-kDrive/pull/1471
Depends on https://github.com/Infomaniak/ios-kDrive/pull/1472

Superseding https://github.com/Infomaniak/ios-kDrive/pull/1310

PR #1444, opened 2025-02-20
fix: Prevent photo library from duplicating items in the upload queue

adrien-coye:
When we re-scan the photo library, we do not exclude what is already in queue.
Therefore, we can end up with multiple uploads of the same file in upload queue.

This PR aims at solving this issue by checking if the hash of the file is queued.
Also, some preventive steps were taken to prevent collisions during the scan of the photo library.

Needed by the 4G feature https://github.com/Infomaniak/ios-kDrive/pull/1440

### DIFF
--- a/kDriveCore/Data/UploadQueue/Servicies/PhotoLibraryUploader+Scan.swift
+++ b/kDriveCore/Data/UploadQueue/Servicies/PhotoLibraryUploader+Scan.swift
@@ -65,8 +65,6 @@ public extension PhotoLibraryUploader {
                     Log.photoLibraryUploader("addImageAssetsToUploadQueue error:\(error)", level: .error)
                 }
             }
-
-            Log.photoLibraryUploader("scheduleNewPicturesForUpload FINISHED")
         }
 
         return newAssetsCount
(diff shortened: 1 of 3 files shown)


kDriveCore/Data/UploadQueue/Servicies/PhotoLibraryUploader+Scan.swift
```diff
@@ -34,36 +34,39 @@ public extension PhotoLibraryUploader {
...
-        let syncDate = Date()
-
-        try? uploadsDatabase.writeTransaction { writableRealm in
-            do {
-                try addImageAssetsToUploadQueue(
-                    assetsFetchResult: assetsFetchResult,
-                    initial: frozenSettings.lastSync.timeIntervalSince1970 == 0,
-                    writableRealm: writableRealm
-                )
+        serialQueue.sync {
```
adrien-coye, line 37: This is not strictly needed, but a defensive step. I'm open to remove it.

adrien-coye (commented):
Left some notes for the reviewer.

kDriveCore/Data/UploadQueue/Servicies/PhotoLibraryUploader+Scan.swift
PhilippeWeidmann: We can keep it

PhilippeWeidmann (approved):

PR #1448, opened 2025-02-24
fix(DriveUser): Remove deprecated avatarUrl

PhilippeWeidmann:


adrien-coye (approved):
I did a migration test. Looking good.

PR #1451, opened 2025-02-25
chore: Add font and mail

Matthieu-dgl:


adrien-coye (commented):
Code and naming looks good. I'll test that it works with the new file types before I approve.

adrien-coye (approved):
Works well, same behaviour as the web. LGTM 👍

PR #1455, opened 2025-02-26
feat: Matomo added for my kSuite

Ambrdctr:


adrien-coye (approved):
I checked every event, I do not see any issue. Approving now 👍

PR #1456, opened 2025-02-27
fix: Use safe on getDisplayedFile

Ambrdctr:


adrien-coye (approved):
Discussed with @Ambrdctr, this fixes a crash but an actual bug persists. 
Since he is actively working on a proper fix I'm merging this for the next beta.

PR #1464, opened 2025-03-04
ci: Use correct path + include sources

PhilippeWeidmann:
Apple doc for ref for $CI_DERIVED_DATA_PATH
https://developer.apple.com/documentation/xcode/environment-variable-reference


adrien-coye (approved):
Looks good, I'll check the behaviour with next release.

PR #1466, opened 2025-03-05
feat: Block only photo sync on wifi only

adrien-coye:
root pr https://github.com/Infomaniak/ios-kDrive/pull/1440

- [x] Test large first synchronisation scaling

This PR is only exists for reference, will be dropped soon.


adrien-coye:
This is not scalling right and easy to maintain. We took another direction here https://github.com/Infomaniak/ios-kDrive/pull/1471

PR #1467, opened 2025-03-06
fix: Realm crash on invalidated object in File action sheet

adrien-coye:
- Reworked `FileActionsFloatingPanelViewController` to function with frozen objects instead of live ones internally.
  - Reworked `setFile()` function to work consistently given any input realm object in any state.
    - Previously, given an entry point, the behaviour could be dissimilar.
- Fix an issue with an invalidated `live` object, since it works with frozen objects internally now.
- In production, it is relatively easy to glitch the "available offline" function. This is now way more robust.

### DIFF
--- a/kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController+Actions.swift
+++ b/kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController+Actions.swift
@@ -74,47 +74,49 @@ extension FileActionsFloatingPanelViewController {
             return
         }
 
-        actions = (frozenFile.isDirectory ? FloatingPanelAction.folderListActions : FloatingPanelAction.listActions).filter { action in
-            switch action {
-            case .openWith:
-                return frozenFile.capabilities.canWrite
-            case .edit:
-                return frozenFile.isOfficeFile && frozenFile.capabilities.canWrite
-            case .manageCategories:
-                return driveFileManager.drive.categoryRights.canPutOnFile && !frozenFile.isDisabled
-            case .favorite:
-                return frozenFile.capabilities.canUseFavorite
-            case .convertToDropbox:
-                return frozenFile.capabilities.canBecomeDropbox
-            case .manageDropbox:
-                return frozenFile.isDropbox
-            case .upsaleColor:
-                return frozenFile.isDirectory && driveFileManager.drive.isFreePack
-            case .folderColor:
-                return frozenFile.capabilities.canColor
-            case .seeFolder:
-                return !normalFolderHierarchy && (frozenFile.parent != nil || frozenFile.parentId != 0)
-            case .offline:
-                return !sharedWithMe && presentationOrigin != .photoList
-            case .download:
-                return frozenFile.capabilities.canRead
-            case .move:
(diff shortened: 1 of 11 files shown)


adrien-coye (commented):
Did a self review, also tested each available action for non regression

adrien-coye:
All actions were checked on files and folder

adrien-coye:
CI is failing for flaky API tests reasons. It builds and test fine.

kDrive/UI/Controller/Files/Rights and Share/ShareAndRightsViewController.swift
```diff
@@ -139,11 +139,12 @@ class ShareAndRightsViewController: UIViewController {
         _ = navigationController?.popViewController(animated: true)
     }
 
-    class func instantiate(driveFileManager: DriveFileManager, file: File) -> ShareAndRightsViewController {
+    class func instantiate(driveFileManager: DriveFileManager, liveFile: File) -> ShareAndRightsViewController {
+        assert(!liveFile.isFrozen && liveFile.realm != nil, "we expect a live object here for this class to work properly")
```
PhilippeWeidmann, line 143: I get a crash here because it's not live -> FileDetailViewController.swift:842

kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController+Actions.swift
```diff
@@ -205,22 +207,26 @@ extension FileActionsFloatingPanelViewController {
     }
 
     private func shareAndRightsAction() {
-        let shareVC = ShareAndRightsViewController.instantiate(driveFileManager: driveFileManager, file: file)
+        guard let liveFile = frozenFile.thaw() else {
+            UIConstants.showSnackBarIfNeeded(error: DriveError.unknownError)
```
PhilippeWeidmann, line 211:
```suggestion
            UIConstants.showSnackBarIfNeeded(error: DriveError.fileNotFound)
```

adrien-coye: Make sense here 👍

PhilippeWeidmann (changes requested):
The new assert in ShareAndRightsViewController.swift:143 crashes the app at least from those locations:
- [ ] FileDetailViewController.swift:842
- [ ] FileListViewModel.swift:372
- [ ] NewFolderViewController.swift:425
- [ ] NewFolderViewController.swift:446
- [ ] NewFolderViewController.swift:488

adrien-coye:
https://github.com/Infomaniak/ios-kDrive/pull/1467#pullrequestreview-2745794249

The crash is an assertion I added in debug only. It shows there is still some access to this method that does not have a correct call pattern. I'll check it right away. This was arguably outside the scope of this PR. (since the assertion is new, not the code)

adrien-coye (commented):
Changed all the calls made to `ShareAndRightsViewController.instantiate` to send a live file as expected.
 did some tests and it all works fine.

PhilippeWeidmann (approved):

PR #1476, opened 2025-03-18
fix: Add LocalizeKit to fix string

Matthieu-dgl:
To check the strings 2 tests: 
- Add a photo to the drive from the Apple photo gallery 
- Add a photo from kDrive


adrien-coye (approved):
I did test this on device and it does fix the issue.

PR #1480, opened 2025-03-21
feat: Add join beta program button

Matthieu-dgl:


adrien-coye (approved):
I did some tests it works well, it matches the spec. Approving now.

PR #1481, opened 2025-03-21
chore: Add matomo event

Matthieu-dgl:


adrien-coye (approved):
There is not an equivalent event on Android apps, therefore a new Matomo event name is fine 👍

PR #1489, opened 2025-03-26
fix: Text preview of large or corrupted files crashes

adrien-coye:
### Abstract

Large text files will crash the app for memory over usage, and corrupted files never complete.

I managed to Swiftify the current code, reduce drastically memory usage. 
But I still had an issue with the JS preview library. So I capped the max size to be previewed.

### DIFF
--- a/kDrive/UI/View/Files/Preview/CodePreviewCollectionViewCell.swift
+++ b/kDrive/UI/View/Files/Preview/CodePreviewCollectionViewCell.swift
@@ -25,12 +25,15 @@ import UIKit
 
 /// Something to read a file outside of the main actor
 struct CodePreviewWorker {
+    /// The JS text preview library will blow up in memory usage if the input is larger
+    static let textFilePreviewCap = 512_000
+
     func readDataToStringInferEncoding(localUrl: URL) async throws -> String {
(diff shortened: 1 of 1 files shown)


kDrive/UI/View/Files/Preview/CodePreviewCollectionViewCell.swift
```diff
@@ -25,16 +25,37 @@ import UIKit
...
+    static let textFilePreviewCap = 512_000
+
     func readDataToStringInferEncoding(localUrl: URL) async throws -> String {
-        let data = try Data(contentsOf: localUrl, options: .alwaysMapped)
-        var maybeString: NSString?
+        let rawData = try Data(contentsOf: localUrl, options: .alwaysMapped)
+
+        let dataToDeserialize: Data
+        if rawData.count > Self.textFilePreviewCap {
+            dataToDeserialize = rawData.prefix(Self.textFilePreviewCap)
```
adrien-coye, line 36: I truncate a memory mapped `Data()`, therefore the entire file is never loaded in memory entirely if too large

PhilippeWeidmann (approved):

PhilippeWeidmann (approved):

PR #1499, opened 2025-04-03
fix(InMemoryFileListViewModel): Crash when adding an existing file to an in memory Realm

adrien-coye:


Ambrdctr (approved):

adrien-coye:
CI was failing as realm migration was impossible (another branch tested on this machine is one version above, known issue)

PR #1512, opened 2025-04-23
fix: OnlyOffice theme is the correct OS theme when editing a document

adrien-coye:
Had to clear the cache when creating a webView to force the correct theme color to be applied.


TommyDL-Infomaniak (approved):

adrien-coye:
Test API failed on a timeout, forcing merge …

PR #1513, opened 2025-04-24
fix: Background refresh can freeze uploads in foreground

adrien-coye:
### Abstract

I found a bug where the `background upload` feature can trigger while the app is running in the foreground. 
This tends to be while the app is running long photo sync uploads.

When the `background upload` is _then_ asked by the system to suspend, the uploads are frozen without the ability to restart without killing the app.

This PR simply reschedule the background task later, and exit gracefully.


adrien-coye:
API tests are failing, not an issue for review.

valentinperignon (approved):

PR #1516, opened 2025-04-29
fix: Scaling issues

adrien-coye:
### Abstract

Some parallel API calls are not limited, here I make sure we do not do more than 4 requests in parallel.


adrien-coye:
I'll test all actions before openning the PR

adrien-coye (commented):
I did a self review and tested all actions modified. All works fine.

valentinperignon (approved):

PR #1518, opened 2025-04-29
feat: Photo sync default to wifi only on new installs

adrien-coye:
New installs defaults to WIFI only for photo sync.
Migration in unchanged.


PhilippeWeidmann (approved):

adrien-coye:
Flaky API tests, force merging now.

PR #1522, opened 2025-05-01
fix: Photo library scan

adrien-coye:

PR #1523, opened 2025-05-02
refactor: Removed photo scan batching

adrien-coye:
#### Abstract

Hashing media assets during photo sync scan can take a long time.
This is performed inside a batched DB transaction. 
This locks the DB in write for the time of a batch of 100 files.
Any write operation on this realm somewhere else will be locking.
Eg. If a user tries to edit sync settings during a batch scan, it can lock the main queue, to a point where the OS will kill the app.

#### Solution
Here I remove completely the DB write batching for the photo sync.
`UploadFile`s are added one by one.

I just make sure the UI correctly `throttle` display and refresh of the views. 
Most screens were already ready for this.
Upload progress is tracked by cell, so it is unchanged by added `throttle` operations
Diffkit calls are unchanged in lists, `throttle` is applied beforehand.

CPU usage is mostly unchanged during scan and uploads.
`main queue` load is reduced as `throttle` is applied in a background queue in the upload screen.

Superseding https://github.com/Infomaniak/ios-kDrive/pull/1522

### DIFF
--- a/kDrive/UI/Controller/Files/File List/FileListViewController.swift
+++ b/kDrive/UI/Controller/Files/File List/FileListViewController.swift
@@ -339,9 +339,12 @@ class FileListViewController: UICollectionViewController, SwipeActionCollectionV
     }
 
     private func bindUploadCardViewModel() {
-        viewModel.uploadViewModel?.$uploadCount.receiveOnMain(store: &bindStore) { [weak self] uploadCount in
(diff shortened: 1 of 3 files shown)


PhilippeWeidmann (approved):

adrien-coye:
Retried CI 14 times to get this to merge

PR #1525, opened 2025-05-06
fix: Empty photo sync upload queue on disable photo sync

adrien-coye:


adrien-coye (commented):
Did a self review.
Also tested that uploads cancel correctly and will restart on re-enabling the feature.
Tested the behaviour with editing photo-sync settings from the "Wi-Fi only" banner. Works seamlessly.

PhilippeWeidmann (approved):

PR #1530, opened 2025-05-08
Fix: Gallery multi selection

Ambrdctr:


adrien-coye (commented):
I just did a review. The code looks good for now.
I found an issue where switching tab while in multiselect will still produce a refresh, and break the UI.

Edit: Like discussed, there is an issue on iOS15

adrien-coye (approved):
The iOS15 issue is fixed. All good for me

PR #1532, opened 2025-05-09
fix: Crash in upload queue view controller

adrien-coye:
#### Abstract

`UploadQueueViewController` had issues working with sync settings from Realm (instead of user defaults).
This PR aims at observing the settings in order to work seamlessly with settings when they get updated.

### DIFF
--- a/kDrive/UI/Controller/Files/Upload/UploadQueueViewController.swift
+++ b/kDrive/UI/Controller/Files/Upload/UploadQueueViewController.swift
@@ -48,7 +48,7 @@ final class UploadQueueViewController: UIViewController {
     var currentDirectory: File!
     private var frozenUploadingFiles = [UploadFileDisplayed]()
     private lazy var sections = buildSections(files: [UploadFileDisplayed]())
-    private var notificationToken: NotificationToken?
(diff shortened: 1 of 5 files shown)


kDrive/UI/Controller/Files/Upload/UploadQueueViewController.swift
```diff
@@ -59,11 +60,36 @@ final class UploadQueueViewController: UIViewController {
         retryButton.accessibilityLabel = KDriveResourcesStrings.Localizable.buttonRetry
         cancelButton.accessibilityLabel = KDriveResourcesStrings.Localizable.buttonCancel
 
-        setUpObserver()
+        setUpFileObserver()
+        setupSyncSettingsObservation()
 
         ReachabilityListener.instance.observeNetworkChange(self) { [weak self] _ in
             Task { @MainActor in
-                self?.reloadCollectionView()
```
adrien-coye, line 66: This was not the correct method to work with our use of diffKit

adrien-coye (commented):
I did a self review and tested that modifying the Wi-Fi only setting was updating the view

PhilippeWeidmann (approved):

Issue #1534, opened 2025-05-11
PDF preview disappears after renaming the file


dixy:
**Description**
The preview of a PDF file disappears ("This file appears to have no preview.") after the file has been renamed while the PDF file is open in preview. The preview still shows the previous file name above the error.

**Steps to reproduce**
1. Open a PDF file
2. Open the option menu at the bottom
3. Click on Rename and change the name 
4. Close the option menu
5. See the PDF preview disappear

**Expected behavior**
The PDF remains visible.

**Smartphone (please complete the following information):**
 - Device: iPad Air M2
 - iOS version: iOS 18.4.1
 - App version: 5.4.0-12

PhilippeWeidmann:
This seems rather easy to reproduce. We check it on our side and update the issue once it's resolved

PR #1538, opened 2025-05-13
fix: Make sure a DB transaction is finished before we notify a change

adrien-coye:
fix https://github.com/Infomaniak/ios-kDrive/issues/1534


adrien-coye (commented):
I did test all actions on a PDF file. It seems OK.

PhilippeWeidmann (approved):

PR #1541, opened 2025-05-15
feat: Add bug report

BaptGrv:


kDriveCore/DI/FactoryService.swift
```diff
@@ -174,6 +175,9 @@ public enum FactoryService {
             },
             Factory(type: MyKSuiteStore.self) { _, _ in
                 MyKSuiteStore()
+            },
+            Factory(type: BugTracker.self) { _, _ in
+                BugTracker(info: BugTrackerInfo(project: "app-mobile-drive"))
```
adrien-coye, line 180: I'm double checking this value.

adrien-coye: We are waiting on a backend change.

PhilippeWeidmann: Bucket is now available `app-mobile-drive` 🎉

adrien-coye (approved):
code LGTM, tested that it does work as expected.

PR #1542, opened 2025-05-15
fix: Update my ksuite usage

lebojo:
Quick update for the usage of my ksuite in kdrive.

Depends on https://github.com/Infomaniak/ios-features/pull/15


adrien-coye (approved):
I did check accounts with and without a mykSuite. 
Did test login and logout.
Did test upgrade from 5.5.0-1
Code LGTM 👍

PR #1551, opened 2025-05-22
fix: Replace floating panel by sheet

lebojo:
Small PR that fixes an issue on modern iPhones or sheets with too many options appearing under the notch.

image


PhilippeWeidmann (commented):
Does it self sizes on larger screens or when there are less options ?

PhilippeWeidmann (approved):

PR #1557, opened 2025-05-27
fix: Add padding to prevent button overlap on table

lebojo:
Small fix to prevent the validation button from going over the folder list in the SelectFolderViewController.


adrien-coye (approved):
Works fine 👍 
I also checked we do not have a `UIConstant.Padding` at 32, so here as a local var it is fine. 
The production version without the SelectViewController (5.4.0) also suffered this issue, so no UI to replicate.

PR #1572, opened 2025-06-06
fix: Folder selection improvements

PhilippeWeidmann:
- Rebuild hierarchy on move
- Start from root when saving file or selecting sync folder for photos
- Prevent showing non writable directories in recent


adrien-coye (approved):
Code lgtm. Works fine with my testing. This goes in the next beta.

PR #1576, opened 2025-06-12
feat: Add slide select in the PhotoListView

BaptGrv:


adrien-coye (commented):
We did a review face to face, code looks good to me.
We can optimise the behaviour of touch recognizers to make sure the tap in only performed if the UIPanGestureRecognizer, I'll DM you some details.

adrien-coye (approved):
I did thoroughly test this PR on iPhone and iPad.
I found some issues that are not linked to this PR, I'll create some bug tickets.
This PR is ready to make it in the next beta 👍

PR #1577, opened 2025-06-12
fix: Handle incompatible videos

lebojo:
Hide play button when video can't be played

This PR fixes an issue where the play button was still shown even if the video couldn't be played. Now, the play button is hidden in such cases.

### DIFF
--- a/kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
+++ b/kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
@@ -486,6 +486,8 @@ final class PreviewViewController: UIViewController, PreviewContentCellDelegate,
             handleOfficePreviewError(error, previewIndex: index)
         } else if file.convertedType == .audio {
             handleAudioPreviewError(error, previewIndex: index)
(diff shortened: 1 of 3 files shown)


kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
```diff
@@ -508,6 +510,17 @@ final class PreviewViewController: UIViewController, PreviewContentCellDelegate,
         downloadFile(at: IndexPath(item: previewIndex, section: 0))
     }
 
+    func handleVideoPreviewError(_ error: Error, previewIndex: Int) {
+        let file = previewFiles[previewIndex]
+
+        guard let avError = error as? VideoPlayer.ErrorDomain,
```
adrien-coye, line 516: This is no longer a cast to an `AVError` from Apple's lib, maybe we can rename it `videoError` ?

adrien-coye (commented):
PR looks good.
Like we discussed, I like the use of `playButton.isEnabled` as it allows us to keep the existing UI with minimal changes.
I'll approve it after I test it in various conditions.

PR #1594, opened 2025-06-27
feat: Preview deeplink without having the file in realm db

BaptGrv:


kDrive/AppRouter+SharedWithMe.swift
```diff
@@ -67,20 +62,31 @@ public extension AppRouter {
...
-                .first?
-                .freezeIfNeeded()
+                                           driveId: Int,
+                                           folderId: Int) async {
+        guard let frozenFolder = await driveFileManager.getFrozenFileFromAPI(
+            driveFileManager: driveFileManager,
+            driveId: driveId,
+            fileId: folderId
+        ) else {
+            return
```
adrien-coye, line 72: Given the recent discussions we had with the iOS team on Monday, I'm starting to think we should have an error snackbar when the deeplink is valid _but_ something fails. I'm asking for validation on this.

kDrive/AppRouter+SharedWithMe.swift
```diff
@@ -34,28 +35,22 @@ public extension AppRouter {
...
-                navigationController: navigationController
-            )
+        guard let presentationOrigin = PresentationOrigin(rawValue: rawPresentationOrigin),
+              let frozenFile = await driveFileManager.getFrozenFileFromAPI(
+                  driveFileManager: driveFileManager,
+                  driveId: driveId,
+                  fileId: fileId
+              )
+        else {
             return
```
adrien-coye, line 49: Same comment here, I'll check if we can / want to show an error if the backend call fails here.

adrien-coye (commented):
I did a review of the code. It does look good to me and is in a good state to be tested.
One questions remains for errors and presenting something to the user if an API call fails. I'm working on validating this. Good job 🙌

adrien-coye (commented):
We are almost there, only a couple of tiny remarks remains.

Lnamw (approved):

PR #1614, opened 2025-07-21
refactor: Use of new method to handle office links

BaptGrv:


kDriveCore/Utils/Routable.swift
```diff
@@ -36,6 +37,8 @@ public enum NavigationRoutes: Equatable {
             return lhdSharedWithMeLink.driveId == rhdSharedWithMeLink.driveId
         case(.trash(let lhdTrashLink), .trash(let rhdTrashLink)):
             return lhdTrashLink.driveId == rhdTrashLink.driveId
+        case(.office(let lhdOfficeLink), .office(let rhdOfficeLink)):
+            return lhdOfficeLink.driveId == rhdOfficeLink.driveId
```
adrien-coye: For comparison to be correct we need to check more than the driveId.
I made `OfficeLink` equatable to check every field in it

PR #1626, opened 2025-08-05
fix: Realm crash access

adrien-coye:
A live realm object, when read to generate an API route in background will crash the app

### DIFF
--- a/kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
+++ b/kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
@@ -530,11 +530,10 @@ final class PreviewViewController: UIViewController, PreviewContentCellDelegate,
     }
 
     func handleOfficePreviewError(_ error: Error, previewIndex: Int) {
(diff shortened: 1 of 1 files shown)


adrien-coye (commented):
I did test this code by forcing an error while displaying an office file, it does still work as intended with a frozen object.

kDriveCore/Data/Cache/PdfPreviewCache.swift
```diff
@@ -66,24 +76,24 @@ public class PdfPreviewCache {
 
                 driveFileManager.apiFetcher.performAuthenticatedRequest(token: token) { token, _ in
                     guard let token else { return }
-                    var urlRequest = URLRequest(url: Endpoint.download(file: file, as: "pdf").url)
+                    var urlRequest = URLRequest(url: Endpoint.download(file: safeFile, as: "pdf").url)
```
adrien-coye, line 79: The crash would be here while building the request from the realm object.

PR #1628, opened 2025-08-05
fix: Realm crash invalidation

adrien-coye:
The local stored realm object could also be invalidated, I also make sure to keep a frozen version.


adrien-coye (commented):
I did test this and download action continues to work with this change

PR #1631, opened 2025-08-06
ci: Try to sleep between API tests to prevent network timeouts

PhilippeWeidmann:


adrien-coye (approved):
Okay, let's try this 👍

PR #1635, opened 2025-08-08
fix: Add new file status

PhilippeWeidmann:


adrien-coye (approved):
I did test that the trash behaviour was OK. Also, I made sure the actions provided are still OK in as sub folder in the trash

PR #1641, opened 2025-08-13
refactor: New method to open privateShareLink

BaptGrv:


adrien-coye (approved):
I'm OK with the code, I like the new functions to split the code.
The /redirect links are working transparently, as they are in production today.
We discussed that we will do the account switching in another PR.

PR #1632, opened 2025-08-06
fix: Move to shared folder support

adrien-coye:
Move to shared folder is now supported as long as we are on the _same_ drive.
I also made sure multiselect was updated too.

PR #1642, opened 2025-08-13
fix: Duplicate bulk action

adrien-coye:
The action was broken since the UI changes of the iPad version

depends on https://github.com/Infomaniak/ios-kDrive/pull/1632


PhilippeWeidmann (approved):

adrien-coye:
Okay, still works fine, merging.

PR #1643, opened 2025-08-13
fix: Modify deeplinks regex to handle all cases

BaptGrv:


adrien-coye (commented):
We can optimise a little somthing.

adrien-coye (approved):
Perfect thanks 👍

PR #1645, opened 2025-08-14
refactor: New method to open directoryLink

BaptGrv:


adrien-coye (approved):
The code looks good, I like to see unit tests 🙌 
I did test that `logged of` then authenticating works _and_ logged in multiple accounts does switch to the correct one. All good to me 👍

PR #1646, opened 2025-08-14
refactor: New method to open filePreview

BaptGrv:


adrien-coye (approved):
Code looks good.
Link works seamless with a simple folder, in various login state.
Let's go 🥳

PR #1647, opened 2025-08-15
fix: Add available offline is always enabled in multi selection

PhilippeWeidmann:


adrien-coye (approved):
Looks good, and works.

PR #1649, opened 2025-08-15
refactor: Replace UniversalLinksHelper with AppRouter+PublicShare

BaptGrv:


adrien-coye (commented):
We did a discuss face to face this.

adrien-coye (approved):
Code looks good.
I did validate the public share behaviour.

PR #1654, opened 2025-08-19
feat: Add a route to handle search deeplinks 

BaptGrv:


adrien-coye (approved):
I did test and verify search, file type, file category, and date interval.
Good work, all works fine 🙌

PR #1656, opened 2025-08-19
fix: Show 'download' and 'offline' for multiselect in offline root

BaptGrv:


adrien-coye (commented):
I did check with Android, we completely hide the move and copy actions within `offlineRoot` in multi-select.

adrien-coye (approved):
I tested this PR this morning with the latest changes, it matches what is done on Android. 
All good to me 👍

PR #1657, opened 2025-08-20
chore: Use `import_loco` with mise and add CI action

valentinperignon:


adrien-coye (approved):
I did test the app on this branch, looks good.
Let's hope the `experimental` flag we use is not going to bite us in the near future, as we discussed.

Issue #1586, opened 2025-06-19
Click on activated links in the new iPad OS sidebar


dixy:
**Is your feature request related to a problem? Please describe.**
With the new sidebar on iPad OS, we cannot click on links already activated in the sidebar.
For example if we click on "Favorites", then click on a folder inside favorites, we cannot click again on "Favorites" in the sidebar to go back to the list of favorites.

**Describe the solution you'd like**
The possibility to click on activated link in the sidebar of the new iPad OS view.

**Describe alternatives you've considered**
Click on another link then go back to the desired link, or click on back, back, back... to go to the start of the navigation (favorites or my folder, ...).

PR #1671, opened 2025-08-28
fix: Can go back to root when clicking on the sidebar element

BaptGrv:
fix: #1586


adrien-coye (commented):
I'll discuss this face to face tomorrow @BaptGrv

adrien-coye (approved):
I re-reviewed it and tested it on iPhone and iPad. It is all good. Good job 🙌

PR #1683, opened 2025-09-09
fix: Handle invalid cursor error

PhilippeWeidmann:


kDriveCore/Data/Cache/DriveFileManager/DriveFileManager+Listing.swift
```diff
@@ -30,7 +30,12 @@ public extension DriveFileManager {
 
         let lastCursor = forceRefresh ? nil : try directory.resolve(within: self).lastCursor
 
-        let result = try await apiFetcher.files(in: directory, advancedListingCursor: lastCursor, sortType: sortType)
+        var result: ValidServerResponse<ListingResult>
```
adrien-coye, line 33: We might want to do this as well in the `public func files(in directory:` in the DriveFileManager.swift as well.

adrien-coye (commented):
Just a comment about something we would add as well

PR #1684, opened 2025-09-09
fix: Changing a right in a shared file is updated live

BaptGrv:


adrien-coye (approved):
Good job, works fine.

PR #1685, opened 2025-09-09
fix: Scroll to top after refreshing last modifications

BaptGrv:


kDrive/UI/Controller/Files/File List/FileListViewController.swift
```diff
@@ -405,6 +405,9 @@ class FileListViewController: UICollectionViewController, SwipeActionCollectionV
             }
         } else {
             refreshControl.endRefreshing()
+            if collectionView.contentOffset.y <= collectionView.contentInset.top {
```
adrien-coye, line 408: I'll make sure it works on _any_ File List

adrien-coye (commented):
Working on QA

kDrive/UI/Controller/Files/File List/FileListViewController.swift
adrien-coye: This was checked. No unwanted side effect were discovered.

adrien-coye:
I leave it to @PhilippeWeidmann to decide if this is a quickwin we want to merge or try to fix the root issue on another PR.

PR #1686, opened 2025-09-10
feat: Early upload start during a photo sync

adrien-coye:
This works since the recent refactors.
- `uploadService.rebuildUploadQueue()` is costly, so I designed something that would not spam it.
- In background mode, we rely on a synchronous execution of the scan then the upload. 
- In background, the app continues to behave like it does today in production.

### DIFF
--- a/kDriveCore/Data/Upload/Servicies/PhotoLibraryUploader/PhotoLibraryUploader+Scan.swift
+++ b/kDriveCore/Data/Upload/Servicies/PhotoLibraryUploader/PhotoLibraryUploader+Scan.swift
@@ -78,6 +78,19 @@ extension PhotoLibraryUploader: PhotoLibraryScanable {
         await worker.value
     }
 
+    private func earlyUploadWhatIsAvailableIfNecessary() async {
+        let filesToUploadCount = uploadDatasource.getAllUploadingFilesFrozen().count
+        let activeUploadsCount = uploadService.operationCount
+
+        guard filesToUploadCount >= Self.startUploadWhileScanningThreshold, activeUploadsCount == 0 else {
+            Log.photoLibraryUploader("No early upload start. Upload state:\(activeUploadsCount)/\(filesToUploadCount)")
+            return
+        }
+
+        Log.photoLibraryUploader("Starting uploads early while scanning")
+        uploadService.rebuildUploadQueue()
+    }
+
     public func cancelScan() async {
         guard let workerTask else {
             return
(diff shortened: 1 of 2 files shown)


kDriveCore/Data/Upload/Servicies/PhotoLibraryUploader/PhotoLibraryUploader+Scan.swift
```diff
@@ -78,6 +79,25 @@ extension PhotoLibraryUploader: PhotoLibraryScanable {
         await worker.value
     }
 
+    private func earlyUploadWhatIsAvailableIfNecessary() async {
+        // The sync _must_ be synchronous in background refresh to work correctly.
```
adrien-coye, line 83: I added a comment to explicit a tricky situation.

kDriveCore/Data/Upload/Servicies/PhotoLibraryUploader/PhotoLibraryUploader+Scan.swift
```diff
@@ -78,6 +79,25 @@ extension PhotoLibraryUploader: PhotoLibraryScanable {
...
+        guard await UIApplication.shared.applicationState == .active else {
```
adrien-coye, line 84: It is a call to UIKit on the main actor I dislike, but until now this is the way the app works to detect current state of refresh.

kDriveCore/Data/Upload/Servicies/PhotoLibraryUploader/PhotoLibraryUploader.swift
```diff
@@ -36,10 +36,13 @@ public protocol PhotoLibraryUploadable {
 public final class PhotoLibraryUploader: PhotoLibraryUploadable {
     @LazyInjectService(customTypeIdentifier: kDriveDBID.uploads) var uploadsDatabase: Transactionable
     @LazyInjectService var uploadService: UploadServiceable
+    @LazyInjectService var uploadDatasource: UploadServiceDataSourceable
 
     /// Threshold value to trigger cleaning of photo roll if enabled
     static let removeAssetsCountThreshold = 10
 
+    static let startUploadWhileScanningThreshold = 1000
```
adrien-coye, line 44: The Threshold is intensionally high in order to not spam `rebuildUploadQueue` which is costly

adrien-coye (commented):
I leave some comments intended for a reviewer

kDriveCore/Data/Upload/Servicies/PhotoLibraryUploader/PhotoLibraryUploader.swift
adrien-coye: Like we discussed, I reduced to 500. 
The upload starts 15 seconds after the scan has started on an iPhone 8.
It is way faster on a modern phone.

PR #1689, opened 2025-09-16
fix: Show correct directory name after restoring from the trash

BaptGrv:


adrien-coye (approved):
Thanks for the changes 👍

PR #1698, opened 2025-09-26
fix: Show floating preview panel only on detailView

BaptGrv:


adrien-coye (approved):
Code looks good.
I did check that it was working OK with other modal presentation like the account detail.
It does work as expected, it will look good with the new revamped iPad menu.

PR #1704, opened 2025-10-01
fix: Disable splitView on iPhone Pro Max

BaptGrv:


kDrive/UI/Controller/Files/SidebarViewController.swift
```diff
@@ -268,7 +268,7 @@ class SidebarViewController: CustomLargeTitleCollectionViewController, SelectSwi
             )
         } ?? []
 
-        if isCompactView && !selectMode {
+        if (isCompactView || UIDevice.current.userInterfaceIdiom == .phone) && !selectMode {
```
adrien-coye, line 271: I think it is safer to add this check `UIDevice.current.userInterfaceIdiom == .phone` _inside_ the computed property `isCompactView`
This way we do not need to change SidebarViewController.swift:325 as well to cover all cases;

adrien-coye (commented):
On SidebarViewController.swift:325 we do not check for interfaceIdiom iphone and we can trigger the iPad view on iPhone pro max.

kDrive/AppRouter.swift
```diff
@@ -464,12 +475,20 @@ public struct AppRouter: AppNavigable {
...
+        if UIDevice.current.userInterfaceIdiom == .pad {
+            let rootSplitViewController = RootSplitViewController(
+                driveFileManager: driveFileManager,
+                selectedIndex: selectedIndex
+            )
+            window.rootViewController = rootSplitViewController
+            window.makeKeyAndVisible()
+            return rootSplitViewController
+        } else {
+            let tabBarViewController = MainTabViewController(driveFileManager: driveFileManager)
```
adrien-coye, line 487: With this change, the apps forgets the selected tab. It used to be passed as a parameter.

```suggestion
            let tabBarViewController = MainTabViewController(driveFileManager: driveFileManager, selectedIndex: selectedIndex)
```

kDrive/AppRouter.swift
adrien-coye, line 473: Since the `window.rootViewController` can now be a `MainTabViewController` this cast no longer work as expected. Maybe use the global `getCurrentController` to do this ?

adrien-coye (approved):
I did test again on iPadOS 26 and 17, looks good.
Small iOS devices are unchanged.
"ProMax" devices does behave like any other iOS devices.
Nice work, this goes into the next beta 🤝

PR #1706, opened 2025-10-03
fix: Add background behind button to hide items

BaptGrv:


adrien-coye (approved):
I did test no regression on iOS. I did test on both iPadOS26 and 17. 
All good to me 👍

PR #1727, opened 2025-11-06
fix: Opening photoSyncSettings via a deeplink on iPad now works

BaptGrv:


adrien-coye (approved):
Cool 👍 
I tested on iPadOS 26 and an older one as well as iOS all works fine.
This goes in the next beta

PR #1737, opened 2025-11-20
fix: Only office content blocker helper

lebojo:
This pull request adds support for displaying a specific error message and FAQ link when a content blocker prevents OnlyOffice documents from loading in the app. It introduces a localized "Read FAQ" string in multiple languages, a new FAQ URL constant, and a custom error handling flow in the `OnlyOfficeViewController`.

WIP: Waiting for new FAQ URL to work


adrien-coye (commented):
I did test this. New error toast works fine.
The user is redirected to the FAQ, which might also be limited with parental control, but they should get a hit at the core issue there.

PR #1740, opened 2025-11-26
fix: Realm crash

adrien-coye:
Fix: When we spawn an in-memory store on the fly I make sure to detach the object we are trying to insert into it.

Rafctor: `InMemoryFileListViewModel` uses a `TransactionExecutor` to make this class behave like the rest of the app.


adrien-coye (commented):
I did test this code with both Trash and Public Share (that uses an In Memory Realm)

PhilippeWeidmann (approved):

PR #1741, opened 2025-11-26
feat: Password protected publicShare

BaptGrv:


kDrive/UI/Controller/Files/PasswordViewController.swift
```diff
@@ -0,0 +1,202 @@
...
+ */
+
+import InfomaniakCoreUIKit
+import InfomaniakDI
+import kDriveCore
+import kDriveResources
+import MaterialOutlinedTextField
+import UIKit
+
+public class PasswordViewController: UIViewController, UITextFieldDelegate {
```
adrien-coye: Cool, this new view is code based and re-uses our existing components.
I'll double-check that the design matches well the Figma later.

kDriveCore/Data/Api/Endpoint+Share.swift
```diff
@@ -44,39 +44,66 @@ public extension Endpoint {
...
+    static func shareLinkAuthentication(driveId: Int, shareLinkUid: String) -> Endpoint {
+        shareUrlV2.appending(path: "/\(driveId)/share/\(shareLinkUid)/auth")
+    }
+
     /// Share link info
-    static func shareLinkInfo(driveId: Int, shareLinkUid: String) -> Endpoint {
-        shareUrlV2.appending(path: "/\(driveId)/share/\(shareLinkUid)/init")
+    static func shareLinkInfo(driveId: Int, shareLinkUid: String, token: String? = nil) -> Endpoint {
+        let shareLink = shareUrlV2.appending(path: "/\(driveId)/share/\(shareLinkUid)/init")
+        if let token {
```
adrien-coye: I think we can factorize this `if let token … shareLink.appending` by adding an extension of `Endpoint`

adrien-coye (commented):
I did test it, code looks good to me.
On iPad and iPhone I think we can make the text input active to automatically scroll at the good level.
ex. `textField.becomeFirstResponder()`
Edit: Check also the behaviour of the `addToMyDrive` action. I had issues in production, this might not be linked to this PR

kDriveCore/Data/Api/Endpoint+Share.swift
```diff
@@ -22,6 +22,14 @@ import RealmSwift
 
 // MARK: - Share Links
 
+private extension Endpoint {
+    func withShareLinkToken(_ token: String?) -> Endpoint {
+        guard let token else { return self }
+        let mergedItems = (queryItems ?? []) + [URLQueryItem(name: "sharelink_password", value: token)]
```
adrien-coye:
```suggestion
        let mergedItems = (queryItems ?? []) + [URLQueryItem(name: PublicShareAPIParameters.password, value: token)]
```

adrien-coye (commented):
As discussed, code reviewed is OK, PR is tested on device and working great.
Only one tiny UX improvement can be added, related to locking the view while it loads.

adrien-coye (approved):
The server has released support for this pr ✅ 

Thanks for the latest changes.
I did test both "single file" and folder public share with password.
If I own the folder, after entering the password, it does redirect to the correct folder.
It all looks good, including multi select.
Standard public share looks unchanged while I tested.
This goes in the next beta, good job 🙌

PR #1743, opened 2025-11-26
fix(SceneDelegate): Only restore app if the app starts

adrien-coye:


adrien-coye (commented):
I did test some more, I see no negative impact on deeplinks or photo sync.

PR #1749, opened 2025-12-09
feat: Add Sentry error when a fileAction is not handled

BaptGrv:


adrien-coye (commented):
Discussed together.

You can set the sentry directly in the `FileActivityType ->  public init(apiRawValue: String)` using a guard.

You can get this way the `apiRawValue` and send it to sentry.

Set the apiRawValue _not_ in the event name, but rather in the `context` in order to make a nice "Dashboard" to track unknown activities.

PR #1750, opened 2025-12-09
feat: Open external link in new tab when creating an account

BaptGrv:


adrien-coye (approved):
Code looks good.
I checked during a register up to the natel validation all the links in the web view and it works nicely.

PR #1754, opened 2025-12-17
fix: Storage progressView color

Lnamw:


kDriveCore/Data/Cache/DriveFileManager/DriveFileManager.swift
```diff
@@ -269,6 +275,17 @@ public final class DriveFileManager {
         return metadata
     }
 
+    public var storageState: StorageState {
+        let storagePercentage = Double(drive.usedSize) / Double(drive.size) * 100
+        if storagePercentage > UIConstants.insufficientStorageMinimumPercentage {
+            return .almostFull
```
adrien-coye: I think this should be .full

PR #1756, opened 2025-12-19
fix: Add debug logs for 2FA early return

PhilippeWeidmann:


adrien-coye (approved):
Code LGTM, I did not test this code.

PR #1769, opened 2026-01-05
fix: Correct selected drive equality

PhilippeWeidmann:
Before
Capture d’écran   2026-01-05 à 10 32 41

After
Simulator Screenshot - iPhone 17 Pro - 2026-01-05 at 10 32 07


adrien-coye (approved):
Ok, discussed why this work. Merging for next Beta.

PR #1771, opened 2026-01-05
feat: Clean local realm files on migration failure

adrien-coye:


kDrive/UI/Controller/Files/File List/InMemoryFileListViewModel.swift
```diff
@@ -57,7 +57,14 @@ class InMemoryFileListViewModel: FileListViewModel {
                 realmAccessible = RealmWrapper(realm: realm)
                 currentDirectory = currentDirectory.detached()
             } catch {
-                Logging.reportRealmOpeningError(error, realmConfiguration: unCachedRealmConfiguration)
+                Logging.reportRealmOpeningError(error, realmConfiguration: unCachedRealmConfiguration, afterRetry: false)
```
adrien-coye: Behaviour of `InMemoryFileListViewModel` is unchanged. I only added code that was moved away from the logger _and_ added a breakpoint in debug.

PR #1772, opened 2026-01-06
feat: Orphan support for kDrive

adrien-coye:
### Abstract

`ApiEnvironment` specific support for `orphans`

depends on https://github.com/Infomaniak/ios-core/pull/178/
depends on https://github.com/Infomaniak/ios-core/pull/179/
depends on https://github.com/Infomaniak/ios-features/pull/57
depends on https://github.com/Infomaniak/ios-features/pull/58

### DIFF
--- a/kDriveCore/Data/Api/Endpoint.swift
+++ b/kDriveCore/Data/Api/Endpoint.swift
@@ -50,6 +50,19 @@ public extension ApiEnvironment {
         }
     }
 
+    var preprodOnOrphanDriveHost: String {
+        switch self {
+        case .prod, .preprod:
+            return "api.\(driveHost)"
+        case .customHost(let host):
+            if host.contains("orphan") {
+                return "api.\(Self.preprod.driveHost)"
+            }
+
+            return "api.\(driveHost)"
+        }
+    }
+
     var onlyOfficeDocumentServerHost: String {
         switch self {
         case .prod, .preprod:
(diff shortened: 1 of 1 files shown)


kDriveCore/Data/Api/Endpoint.swift
```diff
@@ -42,8 +70,10 @@ public extension ApiEnvironment {
             return "info-mq.infomaniak.com"
         case .preprod:
             return "preprod-info-mq.infomaniak.com"
-        case .customHost:
-            Logger.general.error("Cannot guess mqttHost for customHost will fallback to preprod")
+        case let .customHost(host):
+            if !host.contains("orphan") {
+                Logger.general.error("Cannot guess mqttHost for arbitrary customHost will fallback to preprod")
```
adrien-coye: I keep the console warning if we specify something custom that is not an orphan.
When running on orphan, I stay on preprod for this to work

kDriveCore/Data/Api/Endpoint.swift
```diff
@@ -25,15 +25,43 @@ import RealmSwift
...
 
     var onlyOfficeDocumentServerHost: String {
-        return "documentserver.\(driveHost)"
+        switch self {
+        case .prod, .preprod:
+            return "documentserver.\(driveHost)"
+        case .customHost(let host):
+            guard host.contains("orphan") else {
+                Logger.general.error("documentserver for arbitrary customHost will likely fail")
+                return "documentserver.\(host)"
```
BaptGrv: Even if we'll probably never encounter this case, `driveHost` is maybe more coherent here ?

PR #1773, opened 2026-01-06
fix: Added an offset for file previews

BaptGrv:


adrien-coye (approved):
I did test on iPhone/iPad, it all looks good to me.

PR #1774, opened 2026-01-09
fix: Remove interactive pop recognizer

BaptGrv:


kDrive/UI/Controller/Files/Preview/PreviewViewController.swift
```diff
@@ -234,15 +173,35 @@ final class PreviewViewController: UIViewController, PreviewContentCellDelegate,
...
 
-    private func setInteractiveRecognizer() {
-        guard let controller = navigationController else { return }
-        popRecognizer = InteractivePopRecognizer(controller: controller)
-        controller.interactivePopGestureRecognizer?.delegate = popRecognizer
-    }
-
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-        navigationController?.setNavigationBarHidden(true, animated: true)
```
PhilippeWeidmann: Maybe we can still use setNavigationBarHidden when going in "fullscreen" instead of using transforms ?

PR #1778, opened 2026-01-12
fix: Add feedback button and open userReport for non staff

Lnamw:


adrien-coye (approved):
Code looks good 👍 
I did test the code on multiple accounts, it all works flawlessly. Good work 🙌

PR #1779, opened 2026-01-12
fix: Open the personal folder if the user is the owner of the publicShare

BaptGrv:


adrien-coye (commented):
Code LGTM,
I do think we can switch to the correct drive and account (if matching) before we present the modal if we open a deeplink from outside of the app.

PR #1791, opened 2026-01-22
fix: Migrate user profiles if needed

PhilippeWeidmann:


kDrive/UI/Controller/UserProfileMigrator.swift
```diff
@@ -0,0 +1,48 @@
...
+import kDriveCore
+
+struct UserProfileMigrator {
+    @InjectService private var tokenStore: TokenStore
+    @InjectService private var accountManager: AccountManageable
+
+    func migrateUserProfilesIfNeeded() async {
+        let allTokens = tokenStore.getAllTokens().values
+        for token in allTokens {
+            guard await accountManager.userProfileStore.getUserProfile(id: token.userId) == nil else {
```
adrien-coye: Maybe we can factorize the content of this for with a `preloadProfileIfNeeded(for: token)` in the accountManager ?

adrien-coye (commented):
Code looks good. I'll give it a go before I merge it

adrien-coye (approved):
I did test share ext and Files.app.
There exists an issue with Files.app in 5.12.0 that is not linked to this PR.
I did also test a migration from the last 5.11.x to check the "account page" glitch was gone, and it works on my side.

PR #1806, opened 2026-01-28
fix: Following changes on core

Lnamw:
This PR fixes issues with UserDefaults caused by the upgrade to ios-core version 18.5.0.
The changes are intentionally minimal and focused on restoring correct behavior.
A larger refactor is planned and will follow in a subsequent PR.

### DIFF
--- a/kDriveCore/DI/FactoryService.swift
+++ b/kDriveCore/DI/FactoryService.swift
@@ -217,7 +217,7 @@ public enum FactoryService {
                 BackgroundTasksService()
             },
             Factory(type: ReviewManageable.self) { _, _ in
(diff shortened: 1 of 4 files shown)


kDriveCore/DI/FactoryService.swift
```diff
@@ -217,7 +217,7 @@ public enum FactoryService {
                 BackgroundTasksService()
             },
             Factory(type: ReviewManageable.self) { _, _ in
-                ReviewManager(userDefaults: UserDefaults.shared)
+                ReviewManager(userDefaults: UserDefaults.shared, actionBeforeFirstReview: 50)
```
adrien-coye: Cool we can customize it from DI 👍

PR #1816, opened 2026-02-04
fix: NavigationBar doesn't show anymore when leaving the menu

BaptGrv:


adrien-coye (approved):
I did check on both iPad and iPhone, all good to me 👍

PR #1817, opened 2026-02-05
fix(PhotoSyncSettings): Scan from year one not 1970 when performing a photo scan from scratch

adrien-coye:


adrien-coye (commented):
I did a test with pictures tagged up to year 1 AD, I also checked the behaviour when we ask for separate folders (by year). All good on my side on iOS15

PhilippeWeidmann (approved):

PR #1820, opened 2026-02-09
ci: Periphery

adrien-coye:
An implementation that should not slow us up, still allow testing the app correctly, while working great with a mature code base.

To be merged post ios15 drop
depends on https://github.com/Infomaniak/ios-kDrive/pull/1715

To test locally ` periphery scan --retain-codable-properties --format json --relative-results --strict`

Issue to resolve regarding disabling periphery in one target fails in another where the call is superfluous.
My research shows it cannot be silenced

### DIFF
--- a/.github/workflows/periphery.yml
+++ b/.github/workflows/periphery.yml
@@ -1,0 +1,32 @@
+name: CI workflow
+
+on: pull_request
+
+concurrency:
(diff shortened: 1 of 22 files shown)


adrien-coye:
Given the structure of the app using extension and selectively importing part of the source files, there is not a clean and useful way to add Periphery as a quick win.

PR #1828, opened 2026-02-17
fix: Use onlyoffice.infomaniak.com to download onlyoffice files

Ambrdctr:


adrien-coye (approved):
Perfect, all good to me.

PR #1831, opened 2026-02-18
fix: Can delete sharedWithMe if created by me

Ambrdctr:


adrien-coye (commented):
Ok I did check this and it works as the action works today from other entry points.
But the snack bar message is incorrect (moved to the bin). I think we can have a detailed message here (removed permanently).

PR #1873, opened 2026-04-08
feat: Update to Sentry 9

Ambrdctr:


adrien-coye (commented):
I'm ok with the code, but it does not build as is.

adrien-coye (approved):
Okay works on my machine.
I'll make sure to check the behaviour during this first beta.

PR #1876, opened 2026-04-10
fix: Opening a public share from another user of the same drive

BaptGrv:


adrien-coye (approved):
I tested on device.
It works as expected, including the fact it does _not_ switch account automatically to allow importing a public share to other accounts.

PR #1877, opened 2026-04-13
feat: Adding delete button to cancel import of multiple images

killian-mathias:


kDrive/UI/View/Files/Upload/UploadTableViewCell.swift
adrien-coye: You can remove this method since we now only use the version with the `cancelButton`

adrien-coye (commented):
I did test it on iPad and iPhone. It works well, good job.
I only would like the removal of dead code before merging this PR.

adrien-coye (approved):
Thanks for the changes. It all looks good to me.

PR #1878, opened 2026-04-13
fix: Avoid over restoration when returning to the foreground

BaptGrv:


adrien-coye (approved):
To make sure we are safe with this fix I did the following tests.
- Tested with the app lock mechanism ON and OFF
- Tested on iPad also with lock mechanism ON and OFF
All good, thanks for the fix 👍

PR #1880, opened 2026-04-14
fix: Drop children if the cursor is invalid on a listing

BaptGrv:


adrien-coye (approved):
Code is OK, to check for potential side effects a test using "Proxyman/localMap" was done.

PR #1884, opened 2026-04-21
chore: CI trigger build

adrien-coye:


adrien-coye:
Did my test, CI still works

PR #1891, opened 2026-05-05
fix: Fixing duplicate activities by adding a limitation of one request

killianmathias:


kDrive/UI/Controller/Files/FileDetailViewController.swift
```diff
@@ -288,6 +290,7 @@ class FileDetailViewController: UIViewController, SceneStateRestorable {
 
     private func fetchNextActivities() {
         activitiesInfo.isLoading = true
+        requestedActivityAPI = true
         Task { [proxyFile = file.proxify()] in
```
adrien-coye: I think you should track the Task with a property inside the class instead. So you no longer have to track manually a bool.

adrien-coye (commented):
I think we can optimise a little something

adrien-coye (approved):
I did test this latest version on my personal natel. All good on my side. Code matches what I was hoping for 👍 
The API _is_ slow but it now works seamlessly.

PR #1892, opened 2026-05-06
fix: Clearing the cache no longer blocks next downloads

BaptGrv:


adrien-coye (approved):
I did check on device, nice fix. It works on my side.

PR #1894, opened 2026-05-07
chore: Improve storage computation

BaptGrv:


kDriveCore/Data/Upload/Servicies/FreeSpace/CacheModel.swift
```diff
@@ -113,18 +113,23 @@ public enum CacheItem {
...
-        var fileSize: UInt64 = 0
         do {
+            let url = URL(fileURLWithPath: path)
+            let resourceValues = try url.resourceValues(forKeys: [.totalFileAllocatedSizeKey, .fileAllocatedSizeKey])
+            if let allocatedSize = resourceValues.totalFileAllocatedSize ?? resourceValues.fileAllocatedSize {
+                return UInt64(allocatedSize)
+            }
+
             let attributes = try FileManager.default.attributesOfItem(atPath: path)
             if let sizeAttribute = attributes[.size] as? NSNumber {
```
adrien-coye, line 124: Since `attributesOfItem` reads metadata if available, you might want to read the attributes first since it is virtually "free" to do so. And use `resourceValues` as a fallback. Do you agree ?

kDrive/UI/Controller/Menu/StorageTableViewController.swift
```diff
@@ -90,17 +90,8 @@ final class StorageTableViewController: UITableViewController {
...
-           let libraryURL = DriveFileManager.constants.appLibraryDirectoryURL {
-            // Image cache is contained within the app library folder
-            let librarySize = CacheItem.fileSystem(url: libraryURL).size
-            let documentSize = CacheItem.fileSystem(url: documentsURL).size
-            let temporarySize = temporaryCacheItem.size
-            appSize = documentSize + librarySize + temporarySize
-        } else {
-            appSize = 0
-        }
+        let appContainerURL = URL(fileURLWithPath: NSHomeDirectory())
```
adrien-coye, line 105: I guess it is OK to report a total usage that might be larger than the sum of the section we allow to clear as long as it is closer to the truth.

adrien-coye (commented):
Looking good, I have a proposition regarding an efficient way to read file size.
Let me know what you think.

kDriveCore/Data/Upload/Servicies/FreeSpace/CacheModel.swift
adrien-coye: Ok discussed. We keep it this way in order to match closely the size announced by Apple in the settings. All good as is 👍

adrien-coye (approved):
I did test on device, it does produce a value closer to what Apple's settings return. Also the refreshing on display in a nice touch. 👍

PR #1895, opened 2026-05-08
refactor: Avoiding double usage of localUrl and localContainerUrl

killianmathias:
The computed url property may change if the object is live, and it is computed multiple times. 
Here by using a let, we both harden the code and make it more efficient.


adrien-coye (approved):
Cool I did test it to check if I had regressions. None was found. LGTM 👍

PR #1897, opened 2026-05-11
refactor: Remove cocoalumberjack

PhilippeWeidmann:
Used the following logic to replace Cocoalumberjack:
If log line was present multiple times with [SomeClass] -> static logger at top of the class
Else use one of the global logger in Logger+Extension.swift


adrien-coye (approved):
Works on my Natel ™

PR #1898, opened 2026-05-11
refactor: Remove Kingfisher

PhilippeWeidmann:


adrien-coye:
I added a fix to clear the old image cache folders. It is called at the explicit clean-up of image cache _and_ during the existing cache purge mechanism when the app stops.

adrien-coye (approved):
I did thoroughly test it and included a fix to make sure we would not leave a lot of old cache content behind.

PR #1901, opened 2026-05-11
refactor: Cleanup chunking mechanism

adrien-coye:
depends on https://github.com/Infomaniak/ios-kDrive/pull/1900


adrien-coye:
Will need proper re-architecture later

PR #1902, opened 2026-05-11
chore: Add accessibilityLabel and hide image accessibility when not necessary

BaptGrv:


killianmathias (approved):
Looks good to me. I have checked all accessibility labels on RocketSim, and it works well.

PR #1907, opened 2026-05-19
feat: Add menu when long press in FileList

Muuushuuu:
Add a menu when long-pressing on elements
Multi-select buttons and quick actions


kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
```diff
@@ -572,22 +577,67 @@ class FileListViewController: UICollectionViewController, SceneStateRestorable {
...
-                }
+        multipleSelectionViewModel.isMultipleSelectionEnabled = true
+        // Necessary for events to trigger in the right order
+        Task { @MainActor [weak self] in
+            guard let self else { return }
+            if let file = getDisplayedFile(at: indexPath) {
+                multipleSelectionViewModel.didSelectFile(file, at: indexPath)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
Muuushuuu: To fix this issue, I removed the call to `handleLongPress` from `addGestureRecognizer`.
Now, the `handleLongPress` function will only be called by the UIAction in the menu with`IndexPath` parameters to select the good file.

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
```diff
@@ -572,22 +577,67 @@ class FileListViewController: UICollectionViewController, SceneStateRestorable {
...
+                image: action.image,
+            ) { _ in
+                self.viewModel.didMenuAction(
+                    action,
+                    at: indexPath,
+                    presentingParent: self,
+                    sourceView: self.collectionView.cellForItem(at: indexPath) ?? self.collectionView
+                )
             }
         }
```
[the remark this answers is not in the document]
Muuushuuu: Adds the `setupQuickActions` function to select the correct action list. Also adds an “Add” action (for folders) similar to the panel

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
```diff
@@ -32,6 +32,11 @@ class FileListViewController: UICollectionViewController, SceneStateRestorable {
     @LazyInjectService var matomo: MatomoUtils
     @LazyInjectService var accountManager: AccountManageable
     @LazyInjectService var router: AppNavigable
+    @LazyInjectService var downloadQueue: DownloadQueueable
+    #if !ISEXTENSION
+    var downloadAction: FloatingPanelAction?
```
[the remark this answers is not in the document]
Muuushuuu: Removing `downloadAction` served no purpose here

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
```diff
@@ -32,6 +32,9 @@ class FileListViewController: UICollectionViewController, SceneStateRestorable {
     @LazyInjectService var matomo: MatomoUtils
     @LazyInjectService var accountManager: AccountManageable
     @LazyInjectService var router: AppNavigable
+    @LazyInjectService var downloadQueue: DownloadQueueable
+
+    var downloadObserver: ObservationToken?
```
PhilippeWeidmann: This is unused

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
```diff
@@ -572,22 +572,107 @@ class FileListViewController: UICollectionViewController, SceneStateRestorable {
...
-                if let file = getDisplayedFile(at: indexPath) {
-                    multipleSelectionViewModel.didSelectFile(file, at: indexPath)
+        multipleSelectionViewModel.isMultipleSelectionEnabled = true
+        // Necessary for events to trigger in the right order
+        Task { @MainActor [weak self] in
+            guard let self else { return }
+            if let file = getDisplayedFile(at: indexPath) {
+                multipleSelectionViewModel.didSelectFile(file, at: indexPath)
+            }
+        }
```
PhilippeWeidmann, line 584: Check if we still need Task

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
```diff
@@ -572,22 +572,107 @@ class FileListViewController: UICollectionViewController, SceneStateRestorable {
...
+            quickActions = FloatingPanelAction.folderQuickActions
+        } else {
+            quickActions = FloatingPanelAction.quickActions
+        }
+
+        for action in quickActions {
+            switch action {
+            case .shareAndRights:
+                if !frozenFile.capabilities.canShare || offline {
+                    action.isEnabled = false
```
PhilippeWeidmann: `action.isEnabled = false` does nothing because it is never taken into account when you do 
```swift 
UIAction(title: action.name, image: action.image )
```

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
```diff
@@ -572,22 +572,107 @@ class FileListViewController: UICollectionViewController, SceneStateRestorable {
 
     // MARK: - Actions
 
-    @objc func handleLongPress(_ sender: UILongPressGestureRecognizer) {
+    @objc func handleLongPress(indexPath: IndexPath) {
```
BaptGrv: you don't need to expose this function to Objective-C anymore

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
Muuushuuu: I use both variables in the downloadFile method of the FileListViewModel

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
Muuushuuu: It's still necessary, because without it, we have a UI issue. If we switch to multi-selection, we'll see the message “1 file selected” (which is correct), but visually, we can't see which item is selected in our file list (no checkmark).

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
```diff
@@ -572,22 +572,107 @@ class FileListViewController: UICollectionViewController, SceneStateRestorable {
...
+            guard let self else { return }
+            if let file = getDisplayedFile(at: indexPath) {
+                multipleSelectionViewModel.didSelectFile(file, at: indexPath)
+            }
+        }
+    }
+
+    func makeContextMenuQuickActions(indexPath: IndexPath) -> [UIAction] {
+        #if !ISEXTENSION
+        let actions = setupQuickActions(frozenFile: viewModel.getFile(at: indexPath))
```
BaptGrv: you can just do something like this to avoid duplicating `setupQuickActions` function : 

```suggestion
 let file = displayedFiles[indexPath.row]
        let fileInformationsViewController = FileActionsFloatingPanelViewController(frozenFile: file,
                                                                                    driveFileManager: driveFileManager)
        let actions = fileInformationsViewController.quickActions
```

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
```diff
@@ -572,22 +572,107 @@ class FileListViewController: UICollectionViewController, SceneStateRestorable {
...
+        guard !actions.isEmpty else { return [] }
+        return actions.map { action in
+            UIAction(
+                title: action.name,
+                image: action.image,
+            ) { _ in
+                self.viewModel.didMenuAction(
```
BaptGrv: Same here, we can use `fileInformationsViewController.handleAction(action, at: indexPath)` to avoid duplicating `handleAction`

BaptGrv (changes requested):

adrien-coye (commented):
If this PR gets to a state where reviewers are fine with mergin it I'm happy to include it in next Beta.

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
PhilippeWeidmann: Those symbols are in the FileListViewController

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
```diff
@@ -572,24 +569,78 @@ class FileListViewController: UICollectionViewController, SceneStateRestorable {
...
+                attributes: action.isEnabled ? [] : .disabled,
+            ) { _ in
+                fileInformationsViewController.handleAction(action, at: indexPath, isFromMenu: true)
+            }
+        }
+        #endif
+        return []
+    }
+
+    override func collectionView(
```
BaptGrv, line 611: We don't want this long press behavior in the extension

kDrive/UI/Controller/Files/FileActionsFloatingPanelViewController.swift
```diff
@@ -238,12 +241,14 @@ final class FileActionsFloatingPanelViewController: UICollectionViewController {
...
-    func copyShareLinkToPasteboard(from indexPath: IndexPath, link: String) {
+    func copyShareLinkToPasteboard(from indexPath: IndexPath, link: String, isFromMenu: Bool) {
+        let localSourceView = sourceView ?? collectionView.cellForItem(at: indexPath) ?? collectionView
+        guard let localSourceView else { return }
         UIConstants.presentLinkPreviewForFile(
             frozenFile,
             link: link,
-            from: self,
-            sourceView: collectionView.cellForItem(at: indexPath) ?? collectionView
+            from: isFromMenu ? presentingParent! : self,
```
BaptGrv: i would prefer to provide a default value rather than a force-unwrap

BaptGrv (changes requested):

BaptGrv (approved):

PR #1914, opened 2026-05-26
chore: Update tuist to 4.195.6

adrien-coye:
`Tuist` needs to be updated regularly, I waited for the final build of iOS15, which is now behind us, so here is the update PR.

depends on https://github.com/Infomaniak/ios-kDrive/pull/1915


valentinperignon (approved):

adrien-coye (commented):
I did re-test this myself before merging it. It seems to at least build and run with the latest tuist. Beta will tell us of something is broken with the libraries like it was the case once in the past with a broken Tuist version.

PR #1917, opened 2026-05-27
fix: Temporarily disable app uninstall for UI tests

PhilippeWeidmann:


adrien-coye (approved):
This should indeed prevent the booted iOS sim to lock itself.

PR #1918, opened 2026-05-27
fix: Update state of FileListBarButtonType in MultiSelection

Muuushuuu:
When a multi-select event happens, we retrieve the directoryCount from Realm (passed from FileListViewModel)
Then we check directoryCount, if the network call succeeds.
Add a new method that checks the button's state whenever there's a change (Select, Deselect, AllSelect, and AllDeselect)

Depends On #1907


adrien-coye (approved):
Okay so I did test this PR on folders with a lot of files, also while uploads where in progress to check the realm update mechanism. It was slow with a debugger but did not reveal new issues.

PR #1919, opened 2026-05-27
fix: Correct empty view, search bar, and back button animation display

killianmathias:


adrien-coye (approved):
Code looks good, I did test it on a real device. Looking good.

PR #1920, opened 2026-05-28
fix: Correct grid layout in search and hide header for recent searches

killianmathias:


adrien-coye (approved):
I just did test the code on device. LGTM 👍

PR #1923, opened 2026-05-29
feat: Use AppLock from Features library

killianmathias:


kDriveCore/DI/CoreTargetAssembly.swift
```diff
@@ -164,7 +165,11 @@ open class CoreTargetAssembly: TargetAssembly {
                 BGTaskScheduler.shared
             },
             Factory(type: AppLockHelper.self) { _, _ in
-                AppLockHelper()
+                AppLockHelper(
```
adrien-coye, line 168: It looks like it is not building right now. I think the _current_ AppLockHelper version does not provide the constructor with `userDefaults:` It may just be a `Project.swift` issue.

adrien-coye: It builds, thank you 👍

kDrive/UI/Controller/Files/Save File/SaveFileViewController.swift
```diff
@@ -177,21 +178,6 @@ class SaveFileViewController: UIViewController {
         super.viewWillAppear(animated)
         navigationController?.setInfomaniakAppearanceNavigationBar()
         tableView.reloadData()
-        guard UserDefaults.shared.isAppLockEnabled && appLockHelper.isAppLocked else {
```
adrien-coye, line 180: You import AppLock but do not add code here, when sharing a website it no longer block and ask for auth.

adrien-coye (commented):
I'm OK code wise and the locking works well on app extension mode.
My last comment is related to the UI effect, not in this PR for say.

kDrive/SceneDelegate.swift
```diff
@@ -27,7 +28,7 @@ import VersionChecker
 
 final class SceneDelegate: UIResponder, UIWindowSceneDelegate, AccountManagerDelegate {
     @LazyInjectService private var matomo: MatomoUtils
-    @LazyInjectService var lockHelper: AppLockHelper
+    @InjectService var lockHelper: AppLockHelper
```
adrien-coye: Here is SceneDelegate I would prefer we keep it Lazy, as if DI is forced to load early in this Class it may be an issue.

Tuist/Package.resolved
```diff
@@ -87,8 +87,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/Infomaniak/ios-core-ui",
       "state" : {
-        "revision" : "8f50d7bcba8e926be5af324b32be0c1e5d70d6a2",
-        "version" : "24.11.0"
+        "branch" : "feat/remove-app-lock",
```
adrien-coye: I'll give the final approval once this is lifted.

adrien-coye (commented):
I did test the PR, all good to me, ~~I have one small detail regarding the limitations of DI~~.
I'll give the final GO once the dependencies are published (with an SPM TAG /version)

adrien-coye (approved):
I did check on my personal device, as well as on iPad, all good to me 👍

PR #1926, opened 2026-06-03
fix: Disable folder selected while moving

Ambrdctr:


adrien-coye:
As discussed there is an extra modification to do to support multi-select drag-drop

adrien-coye (approved):
This is working well, and tested on my device. This goes in the next Beta.
We did notice a limitation with current architecture that is not a super short fix, so a ticket was created.

PR #1929, opened 2026-06-05
fix: keep mediaHelper reference alive in FileListViewController for photo picker delegate

BaptGrv:


adrien-coye (approved):
Perfect, thanks for the fix 🤝 
Works on my device.

PR #1933, opened 2026-06-10
chore: Point to the updated ios-core version that uses Realm 20

adrien-coye:


adrien-coye:
It builds and run on my device

BaptGrv (approved):

PR #1935, opened 2026-06-10
chore: Re-enable Liquid Glass on kDrive

adrien-coye:
Do not merge. For experiment only


adrien-coye:
Actually implemented in https://github.com/Infomaniak/ios-kDrive/pull/1944

PR #1938, opened 2026-06-12
fix: Apply inset directly on the tabBar frame to avoid glitch

BaptGrv:


adrien-coye (approved):
Did check it is a transparent update for non-notched iPhones.
Did check it looks good on iPhone 13. LGTM 👍

PR #1945, opened 2026-06-19
feat: Open storeviewcontroller from deeplink

PhilippeWeidmann:


adrien-coye (approved):
I was able to use it on device.

PR #1947, opened 2026-06-23
fix: Maintenance view (50x) and Switch account (401)

Muuushuuu:


adrien-coye (approved):
Code is clean 👌
Works great, this goes in the next beta 🥳

PR #1948, opened 2026-06-23
fix: Retry a direct upload after a network connection change

BaptGrv:


adrien-coye (commented):
Code looks good, I'll test it tomorrow 👍

adrien-coye:
I did some tests, It does work better, but I think I was able to reproduce the "waiting" issue with a big file using chunking. Right now the `Task {}` added in this PR can interact with other async work, and I'd like to think about this one a little bit more.

adrien-coye (approved):
I did test switching networks on both iOS 26 and 16 with 4g and 5g networking.

PR #1951, opened 2026-06-25
fix: Upload keeping cache

BaptGrv:


adrien-coye (approved):
Tested on iPad iPhone both 16 and 26, all good on my side 🤝

PR #1954, opened 2026-06-29
fix: Can navigate in parent folder while moving

Ambrdctr:


adrien-coye (approved):
Did test it, all good to me.

PR #1956, opened 2026-06-30
fix: UI glitchs

BaptGrv:


adrien-coye (approved):
I did test on iOS26, all glitches are fixed.
iPad OS26, all good as well, no regressions on iPadOS 16.4
Finally did a check on iOS18 I did not find regressions.
Also code LGTM 👍

PR #1957, opened 2026-06-30
fix: Liquid crash

adrien-coye:
`_UIParallaxDimmingView` is the private view UIKit inserts during a `UINavigationController`
push/pop (the "parallax" transition, also driven by the interactive back-swipe). The exception
means UIKit ran that view's initializer twice on the same instance.

## Root cause

`CustomLargeTitleCollectionViewController` (base class of `HomeViewController` and
`SidebarViewController`) fades the large title in/out by **reassigning the navigation bar's
`standardAppearance` / `compactAppearance` on every scroll event** (`scrollViewDidScroll` →
`updateNavigationBarAppearance()`).

During a push/pop, the collection view's content offset settles, so these scroll callbacks fire
**while the transition is still animating**. Before iOS 26 the navigation bar was static during a
transition, so this was tolerated. With Liquid Glass, UIKit reworked the navigation bar /
transition pipeline: swapping the appearance object mid-transition forces the bar to rebuild its
transition context, which re-initializes the already-created `_UIParallaxDimmingView` and traps.

The interactive back-swipe hits the same path because it scrolls both controllers while
transitioning.

## Fix

In `CustomLargeTitleCollectionViewController`:

- **Skip appearance updates while a transition is in flight** — bail out of
  `updateNavigationBarAppearance()` when `navigationController?.transitionCoordinator != nil`.
  This is the crash fix and also covers the back-swipe.
- **Refresh once the transition settles** — reset the cache and reapply in `viewDidAppear`
  (`transitionCoordinator` is `nil` there), so suppressing updates during the transition never
  leaves a stale/outdated bar.
- **Skip redundant reassignments** — only rewrite the appearance when the computed title alpha
  actually changes, avoiding needless churn of the navigation bar on every scroll tick.

## Scope / notes

- Affects only iOS 26's Liquid Glass navigation transition path; behavior on earlier iOS is
  unchanged (the guard is a no-op when no transition is active).
- The title-fade UX is unchanged on a settled screen.

## Testing

On iPad / iOS 26:
- Scroll **Home** and the **Files** sidebar to a mid-fade title position.
- Push into a folder and pop back, including via the **back-swipe** gesture.
- Verify: no crash, and the title color is correct immediately on return.

### DIFF
--- a/kDrive/Utils/CustomLargeTitleCollectionViewController.swift
+++ b/kDrive/Utils/CustomLargeTitleCollectionViewController.swift
@@ -104,8 +104,8 @@ class CustomLargeTitleCollectionViewController: UICollectionViewController {
             newStandardNavigationBarAppearance.titleTextAttributes[.foregroundColor] = titleColor
             newCompactNavigationBarAppearance?.titleTextAttributes[.foregroundColor] = titleColor
 
-            navigationBar.standardAppearance = newStandardNavigationBarAppearance
-            navigationBar.compactAppearance = newCompactNavigationBarAppearance
+            navigationItem.standardAppearance = newStandardNavigationBarAppearance
+            navigationItem.compactAppearance = newCompactNavigationBarAppearance
         }
 
         guard let originalTitle else { return }
--- a/kDrive/Utils/CustomLargeTitleCollectionViewController.swift
+++ b/kDrive/Utils/CustomLargeTitleCollectionViewController.swift
@@ -50,10 +50,7 @@ class CustomLargeTitleCollectionViewController: UICollectionViewController {
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
 
-        // Appearance updates are suppressed while a push/pop transition is in flight (see
-        // updateNavigationBarAppearance) and the navigation bar appearance is shared across the navigation
-        // stack. Force a refresh once the transition has settled so we never leave a stale/outdated bar on
-        // screen. transitionCoordinator is nil here, so applying the appearance is safe.
+        // Refresh once the transition has settled, since updates are skipped during it.
         lastAppliedTitleAlpha = nil
         updateNavigationBarAppearance()
     }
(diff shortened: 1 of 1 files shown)


kDrive/Utils/CustomLargeTitleCollectionViewController.swift
```diff
@@ -72,6 +82,10 @@ class CustomLargeTitleCollectionViewController: UICollectionViewController {
     }
 
     private func updateNavigationBarAppearance() {
+        // Mutating the nav bar appearance mid push/pop re-inits the parallax dimming view on iOS 26
+        // (Liquid Glass) and crashes: "View was already initialized". Skip while transitioning.
+        guard navigationController?.transitionCoordinator == nil else { return }
```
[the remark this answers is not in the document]
adrien-coye: fixed by 829e72e5a588bcc1e93be27bd95c0381a3a4f214

kDrive/Utils/CustomLargeTitleCollectionViewController.swift
```diff
@@ -82,22 +92,31 @@ class CustomLargeTitleCollectionViewController: UICollectionViewController {
...
 
-        newStandardNavigationBarAppearance.titleTextAttributes[.foregroundColor] = titleColor
-        newCompactNavigationBarAppearance?.titleTextAttributes[.foregroundColor] = titleColor
+            newStandardNavigationBarAppearance.titleTextAttributes[.foregroundColor] = titleColor
+            newCompactNavigationBarAppearance?.titleTextAttributes[.foregroundColor] = titleColor
 
-        navigationBar.standardAppearance = newStandardNavigationBarAppearance
-        navigationBar.compactAppearance = newCompactNavigationBarAppearance
+            navigationBar.standardAppearance = newStandardNavigationBarAppearance
+            navigationBar.compactAppearance = newCompactNavigationBarAppearance
```
BaptGrv: We should apply the new appearance only on the navigationItem instead of modifying the global navbar

BaptGrv (changes requested):

kDrive/Utils/CustomLargeTitleCollectionViewController.swift
adrien-coye: @copilot Please do the change asked by @BaptGrv

adrien-coye: Copi did something too complicated. I did the change.

BaptGrv (approved):

PR #1958, opened 2026-06-30
fix: Add folder color to myKsuitePlus

Ambrdctr:


adrien-coye (approved):
I did check with kSuite Plus and kSuite that it was working, also checked the other action "convert as a dropbox" was working as well.

PR #1965, opened 2026-07-08
refactor: Update list to the system radius

BaptGrv:


adrien-coye (commented):
I just finished testing on iOS16 with an iPhone 8. All good to me 🙌

adrien-coye (approved):
Code looks good 👍  
Using "sections" is a clean trick to switch to native design.

PR #1968, opened 2026-07-10
feat: Add organization on file provider when drive is kSuite

Muuushuuu:


adrien-coye (approved):
Code LGTM, did test on device 💯

PR #1973, opened 2026-07-16
fix: Liquid Glass

BaptGrv:


adrien-coye (commented):
I reviewed the code, all good. iPhone iOS 26 is fine 👍

adrien-coye (commented):
iPhone 8 iOS 16 OKAY

adrien-coye (approved):
I did check again on iPad and iPhone on various OS versions (liquid and not liquid)
Works fine 👍  all good for next beta.

PR #1976, opened 2026-07-21
fix: New NSInternalInconsistencyException in 6.0.0 beta

adrien-coye:
## Context

We were seeing crashes with:

> `NSInternalInconsistencyException` — The collection view's data source returned a cell without a reuseIdentifier.

The crash originated in `SelectFolderViewController`, but the root cause was shared across several file-list collection views.

## Root cause

Several `cellForItemAt` implementations returned a bare `UICollectionViewCell()` in their guard/fallback branches. A directly-allocated cell has no `reuseIdentifier`, which UIKit strictly forbids.

`DifferenceKit` keeps the collection view's item counts in sync with our `sections` model in the normal flow, so these branches were assumed unreachable. But for **off-window** collection views (and on the `interrupt` path), `DifferenceKit` bypasses animated batch updates and calls `reloadData()`. `SelectFolderViewController` pushes a whole stack of controllers at once, so several are off-window. During that `reloadData()`, while the layout is still flushing, the data-source lookup can transiently return `nil` for an index path UIKit is querying → the fallback fabricated an unregistered cell → crash.

## Fix

Never return a non-dequeued cell. Every fallback now returns a dequeued, registered cell instead of `UICollectionViewCell()`:

- **`SelectFolderViewController`** — defers to the base class dequeue and only applies its select-mode customization when the item is an actual file cell (also removes a fragile `as!` force cast).
- **`FileListViewController`** — dequeues a `FileCollectionViewCell` in both fallback branches.
- **`FileCollectionViewCell`** (inner category badges) — dequeues its badge cell before the view-model guard.

The transient inconsistency is now harmless and self-corrects on the next reload instead of crashing.

## Scope / notes

- Audited the whole app: no remaining non-dequeued **cell** returns.
- Supplementary-view fallbacks (`UICollectionReusableView()`) are intentionally left as-is — they are only reached for unregistered kinds where dequeuing is not possible, and no current layout requests them.

## Testing

- Navigate the "Select folder" / move flow through nested folders (including deep-linking into a pre-built navigation stack).
- Verify no crash and cells render correctly on refresh.

### DIFF
--- a/kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
+++ b/kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
@@ -819,7 +819,7 @@ extension FileListViewController {
 
     override func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
         guard let item = sections[safe: indexPath.section]?.elements[safe: indexPath.item] else {
-            return UICollectionViewCell()
+            return collectionView.dequeueReusableCell(type: FileCollectionViewCell.self, for: indexPath)
         }
 
         switch item {
@@ -834,7 +834,7 @@ extension FileListViewController {
             cell.configure(driveFileManager: driveFileManager, presenter: self)
             return cell
             #else
-            return UICollectionViewCell()
+            return collectionView.dequeueReusableCell(type: FileCollectionViewCell.self, for: indexPath)
             #endif
         case .file(let file):
             let cellType: UICollectionViewCell.Type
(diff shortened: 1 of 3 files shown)


kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
BaptGrv: Perhaps here we should avoid force casting and use a guard, as you did in `SelectFolderViewController`

BaptGrv (changes requested):

kDrive/UI/Controller/Files/File List/FileListViewController/FileListViewController.swift
adrien-coye: I managed to not force cast it.

BaptGrv (approved):

PR #1984, opened 2026-07-28
fix: Reduce memory usage in file provider mode

adrien-coye:
I need to asses if this helps or not


adrien-coye:
This approach did not work as I hopped for

PR #1987, opened 2026-08-05
chore: Use SelfSizing helper for custom detents

BaptGrv:


adrien-coye (commented):
Code lgtm, I'm testing on real hardware before I approve.

adrien-coye (approved):
All good on < and >= 26 OS iPhone and iPad

PR #1988, opened 2026-08-06
feat: Deeplinks hardenning

adrien-coye:
## Description

Hardens shared-file deeplinks with a deliberately narrow scope:

> kDrive only accepts regular files placed by trusted Infomaniak apps in the dedicated app-group handoff directory.

## Hardening Methods

- Requires the exact `kdrive-file-sharing://file` route.
- Rejects URLs containing an unexpected path, credentials, port, or fragment.
- Accepts only non-empty `url` query parameters.
- Limits each request to 100 files.
- Requires every supplied path to be absolute.
- Sanitizes filenames using `safeLastPathComponent`.
- Rejects malformed, repeatedly percent-encoded, oversized, or control-character filenames.
- Restricts files to `Library/Caches/file-sharing/<UUID>/<filename>`.
- Uses path components instead of string prefixes for containment checks.
- Resolves symlinks before accepting a file.
- Rejects symlinked handoff directories, UUID directories, and source files.
- Rejects paths escaping the shared app-group handoff directory.
- Accepts regular files only, rejecting directories and other filesystem objects.
- Rejects the complete request when any supplied file is invalid.
- Passes only the validated, resolved file URL to the existing save flow.

## Scope

This hardening applies only to the inter-application file-sharing deeplink. It does not introduce a generic file importer, additional staging, network downloads, or changes to other upload flows.

## Dependency

Requires `ios-core` 19.1.3 for the hardened `safeLastPathComponent` implementation.

depends on https://github.com/Infomaniak/ios-core/pull/205

## Tests

Coverage includes:

- Accepting a regular file from the expected handoff directory.
- Rejecting an incorrect scheme or host.
- Rejecting files outside the handoff directory.
- Rejecting symlink-based path escapes.
- Rejecting directories.


kDriveCore/Utils/DeeplinkParser.swift
```diff
@@ -39,8 +45,14 @@ public struct DeeplinkParser: DeeplinkParsable {
     @LazyInjectService var accountManager: AccountManageable
     @LazyInjectService var router: AppNavigable
 
+    private let sharedContainerURL: URL?
+
     public init() {
-        // META: keep SonarCloud happy
+        sharedContainerURL = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: "group.com.infomaniak")
+    }
```
PhilippeWeidmann: We can drop this constructor and create the path from the TargetAssembly:
`FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: Self.sharedAppGroupName)`

PhilippeWeidmann (changes requested):
One quick change.
Also i wonder if some of this code should be moved in iOS-core in the `DeeplinkService` and the `GroupContainerService` which does the actual writing.

We can share the scheme and other static strings, also protect the writing directly in GroupContainerService

kDriveCore/Utils/DeeplinkParser.swift
```diff
@@ -39,8 +40,16 @@ public struct DeeplinkParser: DeeplinkParsable {
...
+    private let sharedContainerURL: URL?
+
     public init() {
-        // META: keep SonarCloud happy
+        sharedContainerURL = FileManager.default.containerURL(
+            forSecurityApplicationGroupIdentifier: KDriveFileSharing.appGroupIdentifier
+        )
+    }
+
+    init(sharedContainerURL: URL?) {
```
adrien-coye, line 45: This method do not require to be public. It is fine as internal to this repo since we expose a protocol.

kDriveCore/Utils/DeeplinkParser.swift
adrien-coye: changed in https://github.com/Infomaniak/ios-core/pull/205

adrien-coye:
I did check the mail to drive share on device, all works fine

kDriveCore/Utils/DeeplinkParser.swift
```diff
@@ -110,24 +126,83 @@ public struct DeeplinkParser: DeeplinkParsable {
...
+                .standardizedFileURL
+            guard resolvedSourceURL == expectedSourceURL,
+                  let values = try? resolvedSourceURL.resourceValues(forKeys: [.isRegularFileKey]),
+                  values.isRegularFile == true else {
+                return nil
+            }
+
+            files.append(ImportedFile(
+                name: fileName,
+                path: resolvedSourceURL,
```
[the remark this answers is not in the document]
adrien-coye: We consider _our_ sandboxed apps to be safe and do not behave in such a manner withtin the shared space.

PhilippeWeidmann (approved):

PR #1989, opened 2026-08-07
fix: Preview's safe area

BaptGrv:


adrien-coye (approved):
I did check the safe area on mobile and iPad. All good to me.

PR #1990, opened 2026-08-07
refactor: UI test hardening

adrien-coye:


adrien-coye:
this was an experiment, another PR will fix this.

PR #2003, opened 2026-08-21
fix: Update folder name (UI) after get `parentDirectoryId` on Realm

Muuushuuu:
Depends On #2001


adrien-coye (approved):
code looks good 👍

PR #2012, opened 2026-09-03
feat(rca): Add iOS Sentry investigation skill

adrien-coye:


adrien-coye:
The MCP setup will be enough, I'll just make sure it is set up correctly for everyone. Closing this PR

PR #2014, opened 2026-09-03
feat: Add App Entity

BaptGrv:
**Wide-Angle Summary**

PR [#2014](https://github.com/Infomaniak/ios-kDrive/pull/2014) makes kDrive files first-class entities in Apple’s App Intents ecosystem on iOS 18.

In practical terms, it creates a bridge between:

- Apple’s system-level understanding of a “file”
- kDrive’s File Provider integration
- kDrive’s locally cached Realm models
- Existing kDrive operations such as opening, moving, renaming, deleting, and creating folders

This should allow system surfaces that understand Apple’s standard Files intents, such as Shortcuts and other App Intents integrations, to find kDrive files and perform operations on them without reimplementing those operations inside the main UI.

The PR is additive: it introduces 566 lines across 18 files and does not replace an existing workflow.

**What It Adds**

The PR exposes five operations:

- **Open a file**
- **Move one or more files**
- **Rename a file**
- **Delete one or more files**
- **Create a folder**

These use Apple’s predefined Files schemas rather than custom, kDrive-specific intent names:

```swift
@AppIntent(schema: .files.openFile)
@AppIntent(schema: .files.moveFiles)
@AppIntent(schema: .files.renameFile)
@AppIntent(schema: .files.deleteFiles)
@AppIntent(schema: .files.createFolder)
```

That is an important architectural choice: kDrive is presenting itself to the OS as a standard file provider, allowing Apple to understand the meaning and parameters of each operation.

**The Central Abstraction**

The foundation is `KDriveFileEntity` in `kDriveCore/App Intents/KDriveFileEntity.swift:25`.

It is an App Intents `FileEntity` representing either a kDrive file or folder. Alongside its system identifier, it carries enough kDrive-specific information to reconnect the entity to the application:

- User ID
- Drive ID
- File ID
- Internal object UID
- File name
- Creation date
- Modification date

Conceptually, it translates between these two worlds:

```text
Apple FileEntity
      |
      | userId + driveId + fileId
      v
kDrive AccountManager
      |
      v
DriveFileManager
      |
      v
Cached Realm File / kDrive API
```

The entity supports both regular files and folders through `.item` and `.folder` content types.

**How Files Are Identified**

The implementation prefers a real File Provider URL when one is available.

`KDriveFileEntity.makeEntity`:

1. Finds the `NSFileProviderDomain` corresponding to the kDrive.
2. Asks File Provider for the file’s user-visible URL.
3. Creates an Apple `FileEntityIdentifier.file(url:)`.

This gives the system a proper file-backed identity connected to the kDrive File Provider domain.

If File Provider cannot produce a URL, it falls back to a draft identifier:

```text
userId:driveId:fileId
```

That fallback means newly created or not-yet-materialized entities can still be passed between intents even when there is no usable File Provider URL yet.

When an identifier later needs to be resolved, the query reverses this process:

- Draft identifiers are parsed directly into user, drive, and file IDs.
- File URLs are translated back into File Provider item and domain identifiers.
- The domain identifies the kDrive.
- The item identifier identifies the cached file.

This resolution logic is in `KDriveFileEntity.KDriveEntityQuery` at `kDriveCore/App Intents/KDriveFileEntity.swift:117`.

**How System Search Works**

Apple can ask kDrive to find entities matching text, for example while configuring a Shortcut.

The query:

1. Trims and rejects an empty search.
2. Iterates over every locally configured account.
3. Searches every non-maintenance drive belonging to that account.
4. Runs a Realm query for case- and diacritic-insensitive filename matches.
5. Excludes trashed files and internal placeholder IDs.
6. Sorts matches by filename.
7. Returns at most 25 results globally.
8. Converts those matches into File Provider-backed entities where possible.

The latest commit specifically optimizes this path by filtering and limiting in Realm before materializing entities, and by loading File Provider domains once per search rather than once per result.

A significant scope detail is that this is a **local cache search**, not a server-wide kDrive search. Files must be represented in the relevant local Realm database to be discoverable through the intent query.

**How Operations Are Performed**

The intent layer contains very little business logic. It resolves an entity back to a `File` and `DriveFileManager`, checks capabilities, and delegates to existing kDrive services.

The common resolution path is:

```swift
entity.resolveCache()
    -> AccountManager
    -> DriveFileManager
    -> cached File
```

This keeps behavior aligned with operations already used by the application.

### Open

`OpenFileIntent` is slightly different from the mutations because it needs to navigate the application UI.

It:

1. Locates the appropriate `DriveFileManager`.
2. Fetches the requested file.
3. Switches account if the file belongs to another user.
4. Switches drive and reloads the UI if necessary.
5. Shows the main file interface.
6. Presents the selected file using the existing `AppNavigable`/`FilePresenter` flow.

The file is fetched before switching accounts, preventing the account transition from invalidating the context needed to load it.

Relevant code: `kDriveCore/App Intents/KDriveFileIntents.swift:23`.

### Move

The move intent:

- Resolves and validates the destination folder.
- Checks `canMoveInto` on the destination.
- Checks `canMove` on each source.
- Ignores files already inside the selected destination.
- Delegates to the existing `MoveCoordinator`.

Relevant code: `kDriveCore/App Intents/KDriveFileIntents.swift:64`.

### Rename

The rename intent:

- Resolves the target.
- Checks its `canRename` capability.
- Calls `DriveFileManager.rename`.

Relevant code: `kDriveCore/App Intents/KDriveFileIntents.swift:113`.

### Delete

The delete intent:

- Resolves each selected entity.
- Checks its `canDelete` capability.
- Calls the existing deletion API and cache update path.

Relevant code: `kDriveCore/App Intents/KDriveFileIntents.swift:135`.

### Create Folder

The create-folder intent:

- Requests a name if Apple did not supply one.
- Trims and rejects empty names.
- Confirms the destination is a directory.
- Checks `canCreateDirectory`.
- Uses `DriveFileManager.createDirectory`.
- Returns the created folder as another `KDriveFileEntity`, allowing it to feed into subsequent intent steps.

Relevant code: `kDriveCore/App Intents/KDriveFileIntents.swift:158`.

**Security Model**

The mutating operations declare:

```swift
.requiresLocalDeviceAuthentication
```

This applies to move, rename, delete, and create-folder operations. They therefore cannot silently execute while the device is locked and require local device authentication.

Opening a file does not use this explicit mutation policy, because it launches and navigates the application rather than silently changing remote data.

The intents also enforce the capabilities stored on each cached `File`, such as:

- `canMove`
- `canMoveInto`
- `canRename`
- `canDelete`
- `canCreateDirectory`

The server APIs remain the final authority if cached permissions are stale.

**Registration and Packaging**

The actual entities and intents live in `kDriveCore`, alongside the business logic they depend on.

`KDriveCoreIntentsPackage` exposes those definitions:

- `kDriveCore/App Intents/KDriveCoreAppIntentsPackage.swift:21`

The main application then includes that package through:

- `kDrive/App Intents/KDriveExtensionAppIntentsPackage.swift:22`

This separation lets the shared framework own the file behavior while the app target acts as the entry point through which iOS discovers it.

Everything substantive is guarded with `@available(iOS 18.0, *)`, so older supported iOS versions should retain their existing behavior.

**File Provider Change**

The File Provider extension now advertises:

```xml
<key>NSExtensionFileProviderSupportsPickingFolders</key>
<true/>
```

This allows kDrive folders to participate in system folder-selection interfaces, which is necessary for operations such as choosing the destination of a move or selecting where a folder should be created.

**Localization**

The PR adds an `itemLabel` translation in all 13 supported localization files. This is used as the entity type’s display name so Apple’s UI can present a localized label for a kDrive item.

**Overall Architectural Effect**

Before this PR, kDrive files were primarily manipulated through the kDrive UI or the File Provider extension.

After this PR, there is an additional system-facing layer:

```text
Shortcuts / App Intents / Files schemas
                  |
          KDriveFileEntity
                  |
       Account and drive resolution
                  |
      Existing DriveFileManager APIs
                  |
        kDrive API + Realm cache
```

The PR is therefore less about adding new file-management capabilities and more about **making existing kDrive capabilities accessible to iOS through Apple’s standardized file automation model**.

**Current Scope and Caveats**

- It is an iOS 18 feature.
- Search is based on locally cached metadata, not an exhaustive server search.
- Search results are limited to 25.
- Trashed files and drives under maintenance are excluded.
- Multi-file move and delete operations are sequential, not transactional. A later failure could leave earlier items already processed.
- No automated tests were added; Sonar reports 0% coverage on the new code, although its quality gate passed.
- The PR remains open and GitHub indicates that review changes have been requested.

### DIFF
--- a/kDrive/App Intents/KDriveExtensionAppIntentsPackage.swift
+++ b/kDrive/App Intents/KDriveExtensionAppIntentsPackage.swift
@@ -1,0 +1,27 @@
+/*
+ Infomaniak kDrive - iOS App
+ Copyright (C) 2025 Infomaniak Network SA
+
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import AppIntents
+import kDriveCore
+
+@available(iOS 17.0, *)
+struct KDriveAppIntentsPackage: AppIntentsPackage {
+    static var includedPackages: [any AppIntentsPackage.Type] {
+        [KDriveCoreIntentsPackage.self]
+    }
+}
--- a/kDrive/Resources/da.lproj/Localizable.strings
+++ b/kDrive/Resources/da.lproj/Localizable.strings
@@ -1303,6 +1303,9 @@
 /* loco:617b99510ec1ee7d7e2accd2 */
 "inputExpirationTime" = "Udløbstid";
 
+/* loco:6a9a68f38cdf25fae40adb63 */
+"itemLabel" = "Element";
+
 /* loco:6049df4d5c2c3a04bc3979fc */
 "lastEditsTitle" = "Seneste ændringer";
 
--- a/kDrive/Resources/de.lproj/Localizable.strings
+++ b/kDrive/Resources/de.lproj/Localizable.strings
@@ -1303,6 +1303,9 @@
 /* loco:617b99510ec1ee7d7e2accd2 */
 "inputExpirationTime" = "Verfallszeit";
 
+/* loco:6a9a68f38cdf25fae40adb63 */
+"itemLabel" = "Element";
+
 /* loco:6049df4d5c2c3a04bc3979fc */
 "lastEditsTitle" = "Letzte Änderungen";
 
--- a/kDrive/Resources/el.lproj/Localizable.strings
+++ b/kDrive/Resources/el.lproj/Localizable.strings
@@ -1294,6 +1294,9 @@
 /* loco:617b99510ec1ee7d7e2accd2 */
 "inputExpirationTime" = "Ώρα λήξης";
 
+/* loco:6a9a68f38cdf25fae40adb63 */
+"itemLabel" = "Θέμα";
+
 /* loco:6049df4d5c2c3a04bc3979fc */
 "lastEditsTitle" = "Πιο πρόσφατες αλλαγές";
 
--- a/kDrive/Resources/en.lproj/Localizable.strings
+++ b/kDrive/Resources/en.lproj/Localizable.strings
@@ -1294,6 +1294,9 @@
 /* loco:617b99510ec1ee7d7e2accd2 */
 "inputExpirationTime" = "Expiration time";
 
+/* loco:6a9a68f38cdf25fae40adb63 */
+"itemLabel" = "Item";
+
 /* loco:6049df4d5c2c3a04bc3979fc */
 "lastEditsTitle" = "Most recent changes";
 
--- a/kDrive/Resources/es.lproj/Localizable.strings
+++ b/kDrive/Resources/es.lproj/Localizable.strings
@@ -1303,6 +1303,9 @@
 /* loco:617b99510ec1ee7d7e2accd2 */
 "inputExpirationTime" = "Tiempo de expiración";
 
+/* loco:6a9a68f38cdf25fae40adb63 */
+"itemLabel" = "Elemento";
+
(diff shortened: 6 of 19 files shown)


adrien-coye (commented):
Code looks good nice work👍 

I'm happy to see all the abstractions we have been building over the years are leveraged to do this implementation.

adrien-coye (approved):
I did test, all good to me. Good to go in beta.
I openned a ticket for the knitpicks related to the move action I got.

PR #2016, opened 2026-09-07
feat: Spotlight

BaptGrv:


adrien-coye (approved):
I did test the PR, all good to me 👍   This goes into the beta.
I think we can optimize a bit the algo in the next beta, I wrote down a ticket in that regard.

PR #2019, opened 2026-09-08
feat: Semantic index and app intents for iOS27+

BaptGrv:


adrien-coye (commented):
I did test on iPhone 13 pro under iOS 26 build with Xcode 27 without regressions.

adrien-coye (approved):
I was able to test Spotlight indexing on simulator with iOS 27 built on Xcode 27

PR #2024, opened 2026-09-14
fix: Prevent concurrent download observer mutations

adrien-coye:
## Summary

- Fixes `KDRIVE-IOS-1ESH`, an `EXC_BAD_ACCESS` caused by mutating the download observer dictionary while notifications were being published.
- Serializes observer registration and cancellation through a dedicated dispatch queue.
- Publishes notifications from callback snapshots outside the queue, allowing observers to cancel themselves safely without deadlocks.
- Applies the fix to file and archive completion and progress observers.
- Adds regression coverage for cancelling an observer from inside its callback.


adrien-coye (commented):
I did check on device. It works setting 20 ish files in offline mode without issues.

BaptGrv (approved):

PR #2025, opened 2026-09-15
fix: Prevent sending auth token on a publicShare

BaptGrv:


adrien-coye (commented):
I did test public shares on device, still works well 👍

adrien-coye (approved):
I did check that most previews still work as well, including audio streaming.
All good to me 👍

PR #2026, opened 2026-09-15
fix: Add possibility to create a folder on a file import

BaptGrv:


adrien-coye (approved):
I was able to see the button when selecting a folder for photo sync or post scanning a document.

